PR #31, opened 2016-04-18
add_defocus_window

ShakoHo:
@ypwalter  r?


lib/sikuli/gdoc.sikuli/gdoc.py
```diff
@@ -11,9 +11,24 @@ def __init__(self):
         else:
             self.control = Key.CTRL
 
+        if self.os.startswith("M"):
```
ypwalter: how about merging with line 9-12?
if ...:
   self.control = 
   self.alt = 
else:
   self.control = 
   self.alt =

ypwalter:
r+ after neat on insert_image_by_url & alt removal

PR #84, opened 2016-04-26
add-performance-timing-obj-profiler

ShakoHo:
add performance timing obj profiler
@ypwalter r?


ypwalter:
I think maybe we can add comment or readme for lib/perfBaseTest.py
stating the usage of
self.profiler_list.append({"path":"lib.profiler.performanceTimingProfiler", "name":"PerformanceTimingProfiler"})
in setUp() could be helpful

ypwalter:
r+

PR #88, opened 2016-04-28
Fix 2 Profiler class

ypwalter:
Fix 2 Profiler class: (1) Gecko Profiler Bin Fetcher (2) Fix the Old File Name

@ShakoHo r?


ShakoHo:
r+

PR #89, opened 2016-04-28
add-harprofiler

ShakoHo:
@ypwalter r?

### DIFF
--- a/lib/helper/desktopHelper.py
+++ b/lib/helper/desktopHelper.py
@@ -35,6 +35,19 @@ def launch_browser(browser_type, input_profile_path=None):
     else:
         browser_obj = BrowserChrome(DEFAULT_BROWSER_HEIGHT, DEFAULT_BROWSER_WIDTH, profile_path)
     browser_obj.launch()
+    return profile_path
+
+
+def trigger_network_console(browser_type, sikuli_obj, input_env, profiler_list):
+    for profiler_data in profiler_list:
+        if profiler_data['name'] == input_env.PROFILE_NAME_HAR_PROFILER:
+            if browser_type == DEFAULT_BROWSER_TYPE_FIREFOX:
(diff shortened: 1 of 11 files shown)


lib/helper/desktopHelper.py
```diff
@@ -35,6 +35,19 @@ def launch_browser(browser_type, input_profile_path=None):
...
     browser_obj.launch()
+    return profile_path
+
+
+def trigger_network_console(browser_type, sikuli_obj, input_env, profiler_list):
+    for profiler_data in profiler_list:
+        if profiler_data['name'] == input_env.PROFILE_NAME_HAR_PROFILER:
+            if browser_type == DEFAULT_BROWSER_TYPE_FIREFOX:
+                sikuli_obj.run(input_env.sikuli_path, input_env.hasal_dir, "test_firefox_trigger_network",
+                               input_env.test_method_name + "_" + input_env.time_stamp)
```
ypwalter: this should be changed to use file name specified in environment (as we discussed today in the afternoon)

tests/test_chrome_trigger_network.sikuli/test_chrome_trigger_network.py
```diff
@@ -0,0 +1,11 @@
+sys.path.append(sys.argv[2])
+import browser
+import common
+
+
+com = common.General()
```
ypwalter: do you want to remove "import common" and "com = common.General()" if we are not going to use this object here?

tests/test_firefox_trigger_network.sikuli/test_firefox_trigger_network.py
```diff
@@ -0,0 +1,11 @@
+sys.path.append(sys.argv[2])
+import browser
+import common
+
+
+com = common.General()
```
ypwalter: do you want to remove "import common" and "com = common.General()" if we are not going to use this object here?

PR #108, opened 2016-05-19
update readme for opencv installation on mac os x

Conjuror:
there's no need to compile on Mac OS X again and it's hard to build successfully


ShakoHo:
r+

PR #111, opened 2016-05-24
support-pre-run-script

ShakoHo:
@ypwalter @askeing  r?


ypwalter:
is the changes in lib/common/imageTool.py related with supporting prerun scripts?

ShakoHo:
Not fully related to this topic, but we found this issue during testing this patch..so we fix it in the same patch

ypwalter:
There are some "no newline at end of file". Overall, r+.

PR #137, opened 2016-07-04
refactor-sikuli-run-interface

ShakoHo:
@ypwalter r?

### DIFF
--- a/lib/perfBaseTest.py
+++ b/lib/perfBaseTest.py
@@ -53,7 +53,7 @@ class PerfBaseTest(unittest.TestCase):
         self.target_helper = targetHelper.TagetHelper(self.env)
 
         # init sikuli
-        self.sikuli = sikuli.Sikuli()
+        self.sikuli = sikuli.Sikuli(self.env.run_sikulix_cmd_path, self.env.hasal_dir)
 
         # Start video recordings
         self.profilers = Profilers(self.env, self.browser_type, self.sikuli)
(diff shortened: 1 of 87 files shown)


lib/perfBaseTest.py
```diff
@@ -125,7 +125,7 @@ def tearDown(self):
             desktopHelper.stop_browser(self.browser_type, self.env)
 
         # Delete Url
-        if self.test_url_id:
+        if hasattr(self,"test_url_id"):
```
ypwalter: Is there any situation that we might not have test_url_id?

suite.txt
```diff
@@ -1 +1 @@
-test_chrome_gdoc_pd3
+test_firefox_gdoc_read_basic_table_1
```
ypwalter: XD?

lib/perfBaseTest.py
```diff
@@ -53,7 +53,7 @@ def setUp(self):
         self.target_helper = targetHelper.TagetHelper(self.env)
 
         # init sikuli
-        self.sikuli = sikuli.Sikuli()
+        self.sikuli = sikuli.Sikuli(self.env.run_sikulix_cmd_path, self.env.hasal_dir)
```
ypwalter: I didn't see you commit Environment.py in. There is no such default variable currently.

lib/perfBaseTest.py
ShakoHo: yes, when you use pilot testing (given --sikuli_script when executing runtest.py), there will be no test_url_id assign in test script

suite.txt
ShakoHo: XD...no meaning change sorry..

ypwalter:
r+ as it is running smooth in windows now.

PR #144, opened 2016-07-07
support-lock-window-pos

ShakoHo:
@ypwalter @askeing  r?

### DIFF
--- a/lib/common/windowController.py
+++ b/lib/common/windowController.py
@@ -1,11 +1,11 @@
 import time
-import platform
+import sys
 import subprocess
 
-if platform.system().lower() == "windows":
+if sys.platform == "win32":
    import win32gui
    import win32con
-elif platform.system().lower() == "darwin":
+elif sys.platform == "darwin":
     from appscript import *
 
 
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -2,6 +2,7 @@ numpy>=1.10.0
 psutil>=4.1.0
 PyDrive==1.0.1
 pyperclip>=1.5.27
+appscript>=1.01
```
ypwalter: could not found this when installing setup.py scripts in windows.

ypwalter:
not working properly in windows. any prerequisite before runnning in windows?

askeing:
@ypwalter @ShakoHo try this commit https://github.com/Mozilla-TWQA/Hasal/commit/d2aa5d8e5343549176f823cff747ebb8dff4efc0 which add the platform dependencies
(commit on the branch https://github.com/Mozilla-TWQA/Hasal/tree/support-move-desktop-pos-setup )

ypwalter:
r+ for now. will review askeing's pr later for setup.py

requirements_mac.txt
```diff
@@ -0,0 +1 @@
+appscript>=1.01
```
askeing: base on https://pypi.python.org/pypi/appscript, it should be `1.0.1`

askeing: base on https://pypi.python.org/pypi/appscript, it should be `1.0.1`

lib/common/windowController.py
```diff
@@ -0,0 +1,75 @@
+import time
+import sys
+import subprocess
+
+if sys.platform == "win32":
+   import win32gui
+   import win32con
```
askeing: For Windows platform, if we can raise the error message "`please download and install pywin32 from https://sourceforge.net/projects/pywin32/files/pywin32/`" when import modules failed (origin message will be "`ImportError: No module named win32con`"), I think it will be more helpful.

ex:
```
try:
    import win32gui
    import win32con
except ImportError as e:
    print(e)
    raise ImportError('Please download and install pywin32 from https://sourceforge.net/projects/pywin32/files/pywin32/')

ShakoHo: Agree, move the validation to setup.py ..~

PR #145, opened 2016-07-11
Support template matching

MikeLien:
1.remove redundant interface from compare_with_sample_image in imageTool
2.support template matching
  - crop_image supports to crop an image according to tuple list of range, e.g. [(x1,y1), (x2,y2)]
  - template_match calculates comparison results between image and template to find out the most similarity area
  - compare_with_sample_object wraps image list and sample images to compare and find out target results
  - atoi and natural_keys support to get correct result of sorted image sequence names which contain digits

### DIFF
--- a/lib/common/imageTool.py
+++ b/lib/common/imageTool.py
@@ -207,31 +207,24 @@ class ImageTool(object):
             coord = [(0,0), (w,h)]
             print "WARNING: Incorrect coordinates, using fully image crop"
         else:
-            if coord[0][0] < 0:
-                coord[0] = (0, coord[0][1])
-                print "WARNING: Incorrect coordinates, set origin x coordinate to 0"
-            if coord[0][0] > w:
-                coord[0] = (w, coord[0][1])
-                print "WARNING: Incorrect coordinates, set origin x coordinate to width"
-            if coord[1][0] < 0:
-                coord[1] = (0, coord[1][1])
-                print "WARNING: Incorrect coordinates, set target x coordinate to 0"
-            if coord[1][0] > w:
-                coord[1] = (w, coord[1][1])
-                print "WARNING: Incorrect coordinates, set target x coordinate to width"
-            if coord[0][1] < 0:
-                coord[0] = (coord[0][0], 0)
-                print "WARNING: Incorrect coordinates, set origin y coordinate to 0"
-            if coord[0][1] > h:
-                coord[0] = (coord[0][0], h)
-                print "WARNING: Incorrect coordinates, set origin y coordinate to height"
-            if coord[1][1] < 0:
-                coord[1] = (coord[1][0], 0)
-                print "WARNING: Incorrect coordinates, set target y coordinate to 0"
-            if coord[1][1] > h:
-                coord[1] = (coord[1][0], h)
-                print "WARNING: Incorrect coordinates, set target y coordinate to height"
-        print "Crop image range" + str(coord)
+            for i in range(2):
+                for j in range(2):
+                    new_val = coord[i][j]
+                    if coord[i][j] < 0:
+                        new_val = 0
+                    elif j == 0 and coord[i][j] > w:
+                        new_val = w
+                    elif j == 1 and coord[i][j] > h:
+                        new_val = h
+                    if new_val != coord[i][j]:
+                        new_xy = [coord[i][0], coord[i][1]]
+                        list_index = int(j == 0)
+                        new_xy[list_index] = coord[i][list_index]
+                        new_xy[j] = new_val
+                        coord[i] = tuple(new_xy)
+                        print "WARNING: Incorrect coordinates, set %s %s coordinate to %s" % (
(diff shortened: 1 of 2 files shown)


lib/common/imageTool.py
```diff
@@ -124,15 +129,132 @@ def convert_to_dct(self, image_fp):
...
+            if coord[0][1] > h:
+                coord[0] = (coord[0][0], h)
+                print "WARNING: Incorrect coordinates, set origin y coordinate to height"
+            if coord[1][1] < 0:
+                coord[1] = (coord[1][0], 0)
+                print "WARNING: Incorrect coordinates, set target y coordinate to 0"
+            if coord[1][1] > h:
+                coord[1] = (coord[1][0], h)
+                print "WARNING: Incorrect coordinates, set target y coordinate to height"
+        print "Crop image range" + str(coord)
```
ShakoHo: you can simplify the login, for example:

for i in range(2):
    for j in range(2):
        new_val = coord[i][j]
        print coord[i][j]
        if coord[i][j] < 0:
            new_val = 0
        elif j == 0 and coord[i][j] > w:
            new_val = w
        elif j == 1 and coord[i][j] > h:
            new_val = h
        print new_val
        if new_val != coord[i][j]:
            new_xy = [coord[i][0], coord[i][1]]
            list_index = int(j == 0)
            new_xy[list_index] = coord[i][list_index]
            new_xy[j] = new_val
            coord[i] = tuple(new_xy)
            print "WARNING: Incorrect coordinates, set %s %s coordinate to %s" % (["origin", "target"][i],str(unichr(120+j)), str(new_val))

lib/common/imageTool.py
```diff
@@ -24,7 +25,7 @@ def dump_result_to_json(self, data, output_fp):
         with open(output_fp, "wb") as fh:
             json.dump(data, fh)
 
-    def convert_video_to_images(self, input_video_fp, output_image_dir_path, output_image_name=None, exec_timestamp_list=[]):
+    def convert_video_to_images(self, input_video_fp, output_image_dir_path, output_image_name=None, exec_timestamp_list=[], comp_mode=0):
```
askeing: it will be more readable if the default value of `comp_mode` is `False`.

lib/common/imageTool.py
```diff
@@ -51,8 +52,11 @@ def convert_video_to_images(self, input_video_fp, output_image_dir_path, output_
             os.mkdir(output_image_dir_path)
             while result:
                 str_image_fp = os.path.join(output_image_dir_path, "image_%d.jpg" % img_cnt)
-                if (img_cnt >= self.search_range[0] and img_cnt <= self.search_range[1]) or (img_cnt >= self.search_range[2] and img_cnt <= self.search_range[3]):
+                if comp_mode and img_cnt >= self.search_range[0]:
                     cv2.imwrite(str_image_fp, image)
+                else:
+                    if (img_cnt >= self.search_range[0] and img_cnt <= self.search_range[1]) or (img_cnt >= self.search_range[2] and img_cnt <= self.search_range[3]):
+                        cv2.imwrite(str_image_fp, image)
```
askeing: you can use `elif` here.

BTW, what's the diff between `Line 56` and `Line 59`?

lib/common/imageTool.py
MikeLien: comp_mode be True/False is fine to me currently; using default 0 just for consideration of future extension

lib/common/imageTool.py
MikeLien: Both 'Line 56' and 'Line 59' write image out, differences are some images outside search range won't be written

lib/common/imageTool.py
```diff
@@ -124,15 +129,125 @@ def convert_to_dct(self, image_fp):
...
+            sample_fp = os.path.join(input_sample_dp, sample_fn_list[sample_index])
+            if sample_index == 1:
+                print "Template matching will skip the second sample's comparison"
+            elif sample_index == 0:
+                sample_dct = self.convert_to_dct(sample_fp)
+                for img_index in range(self.search_range[1] - 1, self.search_range[0], -1):
+                    image_data = self.image_list[img_index]
+                    comparing_dct = self.convert_to_dct(image_data['image_fp'])
+                    if self.compare_two_images(sample_dct, comparing_dct):
+                        print "Comparing sample file end %s" % time.strftime("%c")
```
askeing: should remove this message due to the comparing still running.

lib/common/imageTool.py
askeing: hmm, if these two condition branches doing the same thing, i'll suggest to merge it into one condition.
ex:
```python
if (comp_mode and img_cnt >= self.search_range[0]) or \
    (img_cnt >= self.search_range[0] and img_cnt <= self.search_range[1]) or \
    (img_cnt >= self.search_range[2] and img_cnt <= self.search_range[3]):
    cv2.imwrite(str_image_fp, image)
```

PR #147, opened 2016-07-13
Refine template matching mechanism

MikeLien:
1. Using threshold to find the first time object shown
2. Output every image which already compared with matching rectangle


ShakoHo:
lgtm r+

PR #150, opened 2016-07-18
Fix close browser failed issue

ShakoHo:
@askeing r?


askeing:
Should remove first 5 commits, only keep latest commit.

PR #154, opened 2016-07-20
Use "open -e" as default editor for Mac OS X

askeing:
I also recommend using `cat` of linux, and `type` of windows for not-blocking reason.


askeing:
@ShakoHo r?

ShakoHo:
lgtm

PR #156, opened 2016-07-21
fix-sikuli-script-name-not-match-video-image-name

ShakoHo:
@askeing @ypwalter r?


ypwalter:
lgtm after I looked through the files changed, I will do a test run later.

PR #157, opened 2016-07-21
Modify default editor of Windows and Linux

askeing:


askeing:
@MikeLien this should also help your pilot tests, r?

PR #169, opened 2016-07-28
Add video profile updater server

askeing:


askeing:
@ShakoHo r?

ShakoHo:
verified in my local env r+

PR #173, opened 2016-08-02
Support only update one of video and profile

### LINKED ISSUE #171: Remove profile_path post data field when uploading video
@askeing

askeing:
Old design only support update video and profile at the same time.
new design will only update video and profile if it has the value

fix #171


askeing:
@ShakoHo r?

ShakoHo:
verified in my local env r+

PR #180, opened 2016-08-05
Modify max-run to 30, and fix the target of URL

askeing:
the `target` of `/os/target/test` is missing the firefox's version info.


askeing:
@ShakoHo r?

PR #185, opened 2016-08-11
Add the 'os/target/' access point of server for getting all target result

askeing:


askeing:
@ShakoHo @ypwalter r?

Issue #213, opened 2016-08-25
Restructure the testname of pilot/reg test cases


askeing:
After having a quick discussion with @ShakoHo , we think the test cases should have the guideline for naming.

We suggest name the cases following the rule `test_<browser>_<webappname>_<case_desc>`.
For example:
* `test_firefox_gmail_view_images`
* all [Facebook pilot test cases]

@ypwalter @MikeLien , do you have any suggestion?
If there is no concern, then we should start to rename cases.

[Facebook pilot test cases]: https://github.com/Mozilla-TWQA/Hasal/tree/master/tests/pilot/facebook

# TODO
- [ ] pilot: BBC
- [x] pilot: facebook
- [x] pilot: gmail
- [ ] reg: gdoc

MikeLien:
I'm fine with it and facebook originally follow this rule to name

PR #214, opened 2016-08-26
Rename Gmail pilot test cases.

askeing:
ref: #213


askeing:
After running all cases on Windows, they work fine.

Issue #218, opened 2016-08-30
Pass the test-tidy


askeing:
Please run following command for checking
```bash
./mach test-tidy --all
```

# 2016-08-30
https://travis-ci.org/Mozilla-TWQA/Hasal/builds/156168767
## TODO
- [x] ./lib/browser/firefox.py:30: no newline at EOF
- [x] ./lib/common/imageTool.py:295: no newline at EOF
- [x] ./lib/common/outlier.py:18: trailing whitespace
- [x] ./lib/common/outlier.py:21: trailing whitespace
- [x] ./lib/profiler/traceloggerProfiler.py:49: no newline at EOF
- [x] ./lib/sikuli/common.sikuli/common.py:2: tab on line
- [x] ./tests/pilot/bbc/test_chrome_editor_pick.sikuli/test_chrome_editor_pick.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_headline.sikuli/test_chrome_headline.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_headline.sikuli/test_chrome_headline.py:10: no newline at EOF
- [x] ./tests/pilot/bbc/test_chrome_homepage.sikuli/test_chrome_homepage.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_homepage_pagedown.sikuli/test_chrome_homepage_pagedown.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_pictures.sikuli/test_chrome_pictures.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_search.sikuli/test_chrome_search.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_search.sikuli/test_chrome_search.py:15: no newline at EOF
- [x] ./tests/pilot/bbc/test_chrome_video.sikuli/test_chrome_video.py:0: the file name should starts with 'test_$BROWSER_bbc_'
- [x] ./tests/pilot/bbc/test_chrome_video.sikuli/test_chrome_video.py:14: no newline at EOF
[...]

PR #221, opened 2016-08-31
fix tidy fails

MikeLien:


MikeLien:
ref #https://github.com/Mozilla-TWQA/Hasal/issues/218

PR #226, opened 2016-08-31
Enable flake8 checking

askeing:
Please run `./mach test-tidy --all` to check the coding style.


MikeLien:
command should be  `./mach test-tidy`

Issue #228, opened 2016-08-31
Fix flake8 error


askeing:
# Steps
1. Fix the error.
2. Run `./mach test-tidy --all` locally to check the result. (Or `./mach test-tidy --all | grep PATH_KEYWORD`)
3. Submit PR with your fix.
4. Merge PR and select your fix in following TODO list.

# Ref
#226 and https://travis-ci.org/Mozilla-TWQA/Hasal/builds/156461389

# TODO
Checking files for tidiness...

## Files
- [x] ./runtest.py
- [x] ./setup.py
- [x] ./lib/perfBaseTest.py
- [x] ./lib/sikuli.py
- [x] ./lib/browser/base.py
- [x] ./lib/browser/chrome.py
- [x] ./lib/common/environment.py
- [x] ./lib/common/imageTool.py
- [x] ./lib/common/outlier.py
- [x] ./lib/common/pyDriveUtil.py
- [x] ./lib/common/recordscreen.py
- [x] ./lib/common/videoFluency.py
- [x] ./lib/common/windowController.py
- [x] ./lib/helper/desktopHelper.py
- [x] ./lib/helper/resultHelper.py
- [x] ./lib/helper/uploadAgentHelper.py
- [x] ./lib/helper/videoHelper.py
- [x] ./lib/profiler/avconvProfiler.py
- [x] ./lib/profiler/base.py
- [x] ./lib/profiler/geckoProfiler.py
- [x] ./lib/profiler/performanceTimingProfiler.py
- [x] ./lib/profiler/traceloggerProfiler.py
- [x] ./lib/sikuli/browser.sikuli/browser.py
- [x] ./lib/sikuli/closeBrowser.sikuli/closeBrowser.py
- [x] ./lib/sikuli/common.sikuli/common.py
- [x] ./lib/sikuli/facebook.sikuli/facebook.py
- [x] ./lib/sikuli/gdoc.sikuli/gdoc.py
- [x] ./tools/HarTransfer.py
- [x] ./tools/get_metric.py
- [x] ./tools/parse_hasal_result.py

## Test Folders
- [x] ./tests/pilot/bbc/
- [x] ./tests/pilot/facebook/
- [x] ./tests/pilot/gmail/
- [x] ./tests/regression/facebook/
- [x] ./tests/regression/gdoc/

askeing:
@ShakoHo @ypwalter @MikeLien 
next challenge XDD

MikeLien:
all errors are fixed

PR #234, opened 2016-09-01
fix helper and profiler flake8 error

MikeLien:


MikeLien:
https://github.com/Mozilla-TWQA/Hasal/issues/228

PR #251, opened 2016-09-06
Add brew ffmpeg cache for Travis CI Mac build

askeing:


askeing:
#252 works fine, so we do not have to cache `ffmpeg` and then link it.

PR #257, opened 2016-09-07
Add checker for checking system packages

askeing:


askeing:
@ShakoHo skip the `Sikuli` system package check on CI, due to there is `Exception in thread "main" java.awt.HeadlessException` when running GUI program on headless machine.

PR #267, opened 2016-09-09
fix real time seq. mismatch real fps

MikeLien:
@ShakoHo r?


MikeLien:
close PR for downgrade FPS setting and set tolerance

PR #268, opened 2016-09-09
fix fps accuracy

MikeLien:
- record setting downgrade
 - get mode of fps from log file


MikeLien:
@askeing r?

MikeLien:
@ShakoHo r?

ShakoHo:
lgtm r+

PR #275, opened 2016-09-14
fix run suite empty line

MikeLien:
- add if statement for valid check
 - reset test_case_module_name to avoid influenced by former case


MikeLien:
@ShakoHo r?

ShakoHo:
Lgtm r+

PR #388, opened 2016-11-09
Updata the links with more generic tags, and remove gmail

### LINKED ISSUE #380: Topsites revisited
Lets track here which topsites need to be added are revisited.

1. **Generic tags/searches**: All things that use `hashtags` or `tags` with `mozilla` (like `https://twitter.com/hashtag/mozilla`) I would rather use a more generic search and something that generates more content, as `mozilla` isn't widely known other than our. I would use more service specific tags, here some shots: `https://twitter.com/hashtag/news`, `https://www.instagram.com/explore/tags/fashion/`

2. **Logins**: `https://mail.google.com/mail/` doesn't work without logins and is also rather complex; could it be removed from the initial set?

askeing:
- fix #380


askeing:
@ShakoHo r?

ShakoHo:
lgtm r+

PR #405, opened 2016-11-14
Found the min timestamp as start time from all tests result

### LINKED ISSUE #403: [Server] Add timestamp at Register for uploading perf data
The different browsers will submit perf data with different timestamp, so we should add a timestamp to sync them.

askeing:
fix #403


askeing:
@ShakoHo r?

ShakoHo:
lgtm r+

PR #420, opened 2016-11-16
Only scroll one way for TopSites scrolling cases

### LINKED ISSUE #418: [TopSites] Scrolling should only move one way
The information from @MikeLien :
If we move two way, and the search range cover the time before and after scrolling, it will match the image before scrolling.

askeing:
Add the methods `scroll_down` and `scroll_up`

fix #418


askeing:
@ShakoHo @MikeLien r?

PR #430, opened 2016-11-17
Fix the ffmpeg issue when PIPE the logging message

### LINKED ISSUE #429: [ffmpeg] Using PIPE will cause the video only record 29~30 seconds

askeing:
fix #429


askeing:
@ShakoHo @MikeLien r?

ShakoHo:
lgtm r+

askeing:
`def stop_recording(self, **kwargs)` the `kwargs` is the interface for caller.

PR #434, opened 2016-11-18
Fix the gecko profiler issue and har profiler issues

askeing:
- and update new GeckoProfilerAutoSaveHAR

```
python runtest.py re <SUITE> --max-run=1 --max-retry=1 --profiler=harexport,geckoprofiler --keep-browser
```


askeing:
@MikeLien r?

PR #448, opened 2016-11-23
Disable Facebook because the Captcha

### LINKED ISSUE #447: [TopSites] Disable Facebook cases
Disable Facebook cases because the Captcha often block the page.
We will get the incorrect result.

askeing:


askeing:
@MikeLien r?

PR #462, opened 2016-11-29
Remove upload resultset to perfherder

ShakoHo:


askeing:
After skimming over the bug, it seems like we should add a mode for local-treeherder for Dev.
Let's open a issue for tracking.

PR #467, opened 2016-12-01
Move the cursor after enter the link on URL bar

askeing:


askeing:
Because I modify the `enterLink()` and `clickBar()` methods, not only `TopSites` but also `regression` cases will be effect.

@MikeLien @ShakoHo r?

ShakoHo:
lgtm r+

PR #473, opened 2016-12-20
Modify the get_metric for easily pasting result data to Sheets

askeing:
Modify the format for easily pasting result data to Sheets.


askeing:
@MikeLien r?

PR #488, opened 2017-01-03
Improve get metrics

ShakoHo:
@askeing r?

### DIFF
--- a/tools/get_metric.py
+++ b/tools/get_metric.py
@@ -6,40 +6,106 @@ from argparse import ArgumentDefaultsHelpFormatter
 
 class ResultMetricGenerator(object):
 
-    def __init__(self, input_result_fp):
+    def __init__(self, input_result_fp, input_mode):
         self.result_fp = input_result_fp
+        self.output_mode = int(input_mode)
+        self.DEFAULT_SHORT_VIEW_MODE = 0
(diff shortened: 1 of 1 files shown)


tools/get_metric.py
```diff
@@ -6,40 +6,102 @@
 
 class ResultMetricGenerator(object):
 
-    def __init__(self, input_result_fp):
+    def __init__(self, input_result_fp, input_mode):
```
askeing: `output_mode` ?

tools/get_metric.py
```diff
@@ -50,8 +112,10 @@ def main():
                                          formatter_class=ArgumentDefaultsHelpFormatter)
     arg_parser.add_argument('-i', action='store', dest='input_result_fp', default=False,
                             help='specify the file need to parse', required=True)
+    arg_parser.add_argument('-m', action='store', dest='input_mode', default=False,
```
askeing: `output_mode` ?

tools/get_metric.py
```diff
@@ -6,40 +6,102 @@
 
 class ResultMetricGenerator(object):
 
-    def __init__(self, input_result_fp):
+    def __init__(self, input_result_fp, output_mode):
```
askeing: `def __init__(self, input_result_fp, output_mode=0):` would be better :P

Issue #493, opened 2017-01-05
 Ensure Hasal test configuration is using recommend benchmarking settings


overholt:
This was filed in Mozilla Bugzilla here: https://bugzilla.mozilla.org/show_bug.cgi?id=1328949

askeing:
Stage 1 finished.
Currently it works for running `--firefox-settings settings/default.json`, and profiling`--firefox-settings settings/default.json --profiler=geckoprofiler`.
For running, it will create a new Firefox profile with `prefs` and `cookies` you want. And then running with `env`.
For profiling, it still is temporary solution. It will put the `prefs` and `cookies` into unzipped profile folder (from `resource/*.zip`). And then running with `env`.

Stage 2 will change the mechanism for installing XPI (addons, extensions), and enabling profiler.

PR #498, opened 2017-01-10
Firefox configuration - part 1: create profile with prefs and cookies

askeing:
Ref: #493


askeing:

runtest.py
```diff
@@ -69,6 +86,63 @@ def clean_up_output_data(self):
...
+
+    def _create_firefox_profile(self):
+        prefs = self.settings_prefs
+        tmp_profile_dir = tempfile.mkdtemp(prefix='firefoxprofile_')
+        self.logger.info('Creating Profile: {}'.format(tmp_profile_dir))
+        self.logger.info('Profile with prefs: {}'.format(prefs))
+        if sys.platform == 'linux2':
+            firefox_cmd = 'firefox'
+            # the command "--profile <PATH> -silent" doesn't work on Ubuntu
+            os.system('{} --profile {} &'.format(firefox_cmd, tmp_profile_dir))
```
ypwalter: Shouldn't we use -createprofile "(name) (folder)" instead?

ypwalter: nice finding by askeing! It can be silent created if the folder is created!

askeing: you should create empty folder first.

By the way, after the folder be created, it can create profile with `-silent`.
I will update the code.

askeing:
Base on @ShakoHo's comment, extract Firefox profile related methods to new class.

askeing:
**IMPORTANT**
The login session will miss when we create new profile for running Firefox.
Do not merge it before we have the solution for it.

ShakoHo:
so far so good, lgtm r+

askeing:
Adding `_download_cookies()` for downloading the prepared Firefox `cookies` from Google Drive.
So that we can run our Google Suite/Facebook performance cases without login actions.

PR #503, opened 2017-01-12
Firefox configuration - part 1.5 Do not default enable loading settings from file

askeing:
Running Firefox with user's default profile by default.
Generating new profile and running Firefox with this new one only if specify the `--firefox-setting`.

Ref: #493


askeing:

lib/helper/firefoxProfileCreator.py
```diff
@@ -72,7 +72,7 @@ def _install_profile_extensions(self, extensions_settings={}):
         return
 
     def remove_firefox_profile(self):
-        if os.path.isdir(self._firefox_profile_path):
+        if self._firefox_profile_path and os.path.isdir(self._firefox_profile_path):
```
ypwalter: I think self._firefox_profile_path was initialed as an empty string. Therefore, os.path.isdir('') will return False and checking of self._firefox_profile_path would be unnecessary?

askeing: it's necessary when `self._firefox_profile_path = 'foo/bar/no/folder'`

runtest.py
```diff
@@ -88,10 +90,17 @@ def clean_up_output_data(self):
...
 
     def _load_settings(self):
+        if not self.firefox_settings:
+            return {}
         try:
             if os.path.exists(self.firefox_settings):
                 with open(self.firefox_settings) as firefox_settings_fh:
-                    return json.load(firefox_settings_fh)
+                    settings = json.load(firefox_settings_fh)
+                    self.logger.info('** Important **\n'
```
ypwalter: Try to use a larger warning slogan like:


           *****************************
           *                WARNING               *
           *****************************
THIS WILL OVERRIDE YOUR CURRENT FIREFOX SETTINGS.
EXIT NOW IF YOU DON'T WISH TO DO SO.


I believe people could see this more easier.

lib/helper/firefoxProfileCreator.py
ypwalter: ok, this would make check faster, but you might want to comment on this for the reason.

runtest.py
askeing: well, I already wait few seconds for notice people.
I think we put the message to logger, and logger will handle the message base on it's format string.

If we want to show larger warning message like:
```
##########
# WARNING #
##########
some message...
```
we have to put `\n` in the beginning of message, that will break the logger's format.

ypwalter: I skipped '\n' part because I wasn't writing the code, but adding a '\n' is totally fine. I just want to make sure people see the warning clearly before they get back to complain this.

ypwalter:
r+

lib/helper/firefoxProfileCreator.py
askeing: In my opinion, it's not only for speed.
The meaning of this line is simpler than only put `isdir` check here, `if we have path, and the path is a folder, then do something`. Do we have to add comment to describe that?

PR #506, opened 2017-01-16
disable firefox block popup

MikeLien:


askeing:
lgtm, r+

note: for Google Slides, presentation mode

PR #507, opened 2017-01-17
Disable the Firefox promt "Firefox automatically sends some data to Mozilla so that we can improve your experience"

askeing:


askeing:
@MikeLien @ShakoHo r?

askeing:
Ref: 
http://stackoverflow.com/questions/28410049/how-to-disable-firefox-promt-firefox-automatically-sends-some-data-to-mozilla-s

PR #520, opened 2017-01-24
Fix the issue of cv2.VideoCapture is not opened

askeing:
ref:
The `vidcap.isOpened()` is `False` if Video file is not ready.
And the `result, image` will be `Fasle, None`.
```
vidcap = cv2.VideoCapture(input_video_fp)
result, image = vidcap.read()
```


askeing:
@ShakoHo @MikeLien r?

MikeLien:
lgtm r+

PR #536, opened 2017-02-14
Add some comment for imageTool

askeing:
1st verion


askeing:
@MikeLien r??????  -.-?

Issue #547, opened 2017-02-17
Input Latency MVP


digitarald:
Measure time from input to screen updated.

@MikeLien, can you fill in the details on what is planned?

MikeLien:
Currently the initial plan is to integrate prototyping case in Hasal framework.
There have some tasks need to be finished in this integration.
- [ ] Add trigger point for input latency of typing case
- [ ] Switch test type to give other indexes
- [ ] Modify analysis code
- [ ] Add specific typing case

digitarald:
@MikeLien can you confirm that this will not only cover typing but any interaction, like clicking a button, focusing a field, etc?

MikeLien:
For getting whole period from input event to display shown, we propose a new solution to get input latency.

1. Snapshot before input command
2. Input event with log
3. Snapshot after display shown

This solution can apply to not only typing case but all interactions with log in console.
- For clicking a button, the latency would be from clicking command shown on console until button display change.
- For focusing a field, the latency would be from focus action/command shown on console until the field is focused and display changed.

bobbyc:
Detail definition of Actual input latency in [Google Docs](https://docs.google.com/document/d/1nZZmcXbbuXRt0ilXTs7c1CPYdiH1bb6kp2Zk8FtywgY/edit)

bobbyc:
Per Mike's test:

Below are experimental results for input latency of clicking from https://codepen.io/digitarald/pen/gmwyde 

Chrome: 752.81 ms
Generated result GIF file for 10 times slower from input event log to website display change: https://drive.google.com/a/mozilla.com/file/d/0B9Zi9TqbRWsdOGV5Sm4xWGZ6a0xabWpLLTZNMVFhR0dRNFZJ/view?usp=sharing

Firefox: 764.04 ms
Generated result GIF file for 10 times slower from input event log to website display change: https://drive.google.com/open?id=0B9Zi9TqbRWsdZWs0VHFYS2VST2dTQ0tTeDFzOHRjdnNuQWZz

digitarald:
@MikeLien what are the numbers reported in the script on the page to compare with? They are not visible in the loop. ~750ms seems way too high, I expect between 30-60ms.

digitarald:
This link opens full screen without iframes: https://s.codepen.io/digitarald/debug/gmwyde/VGrWNnNyoyNM

bobbyc:
@digitarald your latest codepen link is not working.

[...]

PR #559, opened 2017-03-09
Move Terminal window position for AIL

askeing:
@MikeLien @ShakoHo r?

ref: #547

### DIFF
--- a/lib/baseTest.py
+++ b/lib/baseTest.py
@@ -65,14 +65,26 @@ class BaseTest(unittest.TestCase):
                                                self.env.img_output_sample_1_fn)
                     if desktopHelper.check_browser_show_up(self.env.img_sample_dp, self.env.img_output_sample_1_fn):
                         logger.debug("Browser shown, adjust viewport by setting.")
-                        height_adjustment, width_adjustment = desktopHelper.adjust_viewport(self.browser_type,
-                                                                                            self.env.img_sample_dp,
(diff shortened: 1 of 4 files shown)


MikeLien:
local integration test passed, r+

lib/baseTest.py
```diff
@@ -55,8 +68,32 @@ def get_browser_done(self):
...
                     if desktopHelper.check_browser_show_up(self.env.img_sample_dp, self.env.img_output_sample_1_fn):
                         logger.debug("Browser shown, adjust viewport by setting.")
-                        desktopHelper.adjust_viewport(self.browser_type, self.env.img_sample_dp,
-                                                      self.env.img_output_sample_1_fn)
+                        height_browser, width_browser = desktopHelper.adjust_viewport(self.browser_type,
+                                                                                      self.env.img_sample_dp,
+                                                                                      self.env.img_output_sample_1_fn)
+                        height_offset = 0
+                        terminal_width = width_browser
+                        terminal_height = 60
```
ypwalter: not needed when you reset terminal_height in the following if-else statement.

lib/baseTest.py
```diff
@@ -55,8 +68,32 @@ def get_browser_done(self):
...
+                        if sys.platform == 'linux2':
```
ypwalter: 9&45 might be better for rack machines per tested today

PR #566, opened 2017-03-15
Add si generator

ShakoHo:
@MikeLien r?


MikeLien:
LGTM and local test passed, r+

PR #573, opened 2017-03-22
AIL gdoc case extended - a 10 pages case

ypwalter:
@MikeLien r?


MikeLien:
lgtm r+

ypwalter:
verified in multiple OS done.

PR #576, opened 2017-03-22
fix infinite loop when image comparsion can't find anyone of begin an…

ShakoHo:
@MikeLien @askeing r?


lib/helper/generatorHelper.py
```diff
@@ -170,21 +170,24 @@ def output_result(test_method_name, result_data, output_fp, time_list_counter_fp
...
+                comparing_time_data[time_point] = event_data['time_seq']
+                break
 
     event_time_dict = dict()
-    for event_data in current_run_result:
-        for event_name in event_data:
-            if event_name != 'time_seq' and event_name != 'start' and event_name != 'end':
-                event_time_dict[event_name] = np.absolute(event_data['time_seq'] - start_time)
+    if len(comparing_time_data.keys()) == 2:
+        run_time = comparing_time_data['end'] - comparing_time_data['start']
```
MikeLien: suggest add one more condition to judge if run_time is greater than 0 to prevent false results

PR #579, opened 2017-03-22
G-Doc Pagedown Input Latency

askeing:
To test run this PR, you have to modify following two files `lib/baseTest.py`, `lib/common/environment.py` (see below).
```diff
diff --git a/lib/baseTest.py b/lib/baseTest.py
index 22933b7..518d00c 100644
--- a/lib/baseTest.py
+++ b/lib/baseTest.py
@@ -7,7 +7,7 @@ import unittest
 import helper.desktopHelper as desktopHelper
 import lib.helper.videoHelper as videoHelper
 import lib.helper.targetHelper as targetHelper
-import helper.resultHelper as resultHelper
+import helper.generatorHelper as generatorHelper
 from common.environment import Environment
 from common.logConfig import get_logger
 from common.windowController import WindowObject
@@ -147,12 +147,12 @@ class BaseTest(unittest.TestCase):
         # output result
         if self.round_status == 0:
             if hasattr(self, "crop_data"):
-                resultHelper.result_calculation(self.env, self.crop_data,
+                generatorHelper.calculate(self.env, self.crop_data,
                                                 int(os.getenv("CALC_SI")), int(os.getenv("ENABLE_WAVEFORM")),
                                                 os.getenv("PERFHERDER_REVISION"), os.getenv("PERFHERDER_PKG_PLATFORM"),
                                                 os.getenv("SUITE_UPLOAD_DP"))
             else:
-                resultHelper.result_calculation(self.env, calc_si=int(os.getenv("CALC_SI")),
+                generatorHelper.calculate(self.env, calc_si=int(os.getenv("CALC_SI")),
                                                 waveform=int(os.getenv("ENABLE_WAVEFORM")),
                                                 revision=os.getenv("PERFHERDER_REVISION"),
                                                 pkg_platform=os.getenv("PERFHERDER_PKG_PLATFORM"),
```

```diff
diff --git a/lib/common/environment.py b/lib/common/environment.py
index e3d6ec8..b5edfd9 100755
--- a/lib/common/environment.py
+++ b/lib/common/environment.py
@@ -170,8 +170,8 @@ class Environment(object):
         self.video_output_sample_2_fp = os.path.join(self.DEFAULT_VIDEO_OUTPUT_DIR, self.output_name + "_sample_2.mkv")
         self.img_sample_dp = os.path.join(self.DEFAULT_IMAGE_SAMPLE_DIR, self.output_name)
[...]

PR #582, opened 2017-03-23
G-Sheet Pagedown Input Latency

askeing:
ref: #579


askeing:

PR #584, opened 2017-03-23
Rename the AIL pagedown cases name

askeing:
ref: #581


askeing:

PR #585, opened 2017-03-23
Auto mail robot to trigger job on demand

Conjuror:
Propose a solution for #571 

note that the agent requests manual approval for the very first time.
Also, there is a issue that we'll need to install google-api-python-client library on Windows manually now. (should be scriptable)

### DIFF
--- a/agent/iskakalunan.py
+++ b/agent/iskakalunan.py
@@ -40,6 +40,7 @@ LABEL_REPLIED     = 'Label_6'
 RESULT_ZIP = 'result.zip'
 RESULT_JSON = 'result.json'
 UPLOAD_FOLDER = 'upload'
+FIREFOX_PATH = r'C:/Program Files/Mozilla Firefox'
 
 MSG_TYPE_INVALID_REQUEST = 'This is a notifier that your request is invalid.'
 MSG_TYPE_UNDER_EXECUTION = 'This is a notifier that your request is under execution.'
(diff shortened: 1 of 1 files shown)


Conjuror:
I just realized that we have google-api-python-client 1.5.3 on the machines. I'll check if we need to upgrade to 1.6.2 for auto mail robot.

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+        if status == MSG_TYPE_JOB_FINISHED:
+            path = os.path.join(os.getcwd(), RESULT_ZIP)
+            content_type, encoding = mimetypes.guess_type(path)
+
+            if content_type is None or encoding is not None:
+                content_type = 'application/octet-stream'
+            main_type, sub_type = content_type.split('/', 1)
+            with open(path, 'rb') as f:
+                msg = MIMEBase(main_type, sub_type)
+                msg.set_payload(fp.read())
```
ShakoHo: fp? f? typo?

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+            return True
+        else:
+            logger.info("Job did not finish by checking")
+            return False
+
+    def reply_status(self, msg_src, status):
+        """Update status to sender"""
+        logger.debug("reply status to %s with %s" % (msg_src.message_id, status))
+        message = MIMEMultipart()
+        message['From'] = 'moztpeqa@mozilla.com'
```
ShakoHo: Should we consider to put email address as a global variable?  The email address modification could happen in nearly future

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+            while 'nextPageToken' in response:
+                page_token = response['nextPageToken']
+                response = service.users().messages().list(userId=user_id,
+                                                           labelIds=label_id,
+                                                           pageToken=page_token).execute()
+                message_list.extend(response['messages'])
+
+            for message in message_list:
+                message_temp = Message(self.service, message['id'], message['threadId'])
+                if not 'moztpeqa' in message_temp.sender:
```
ShakoHo: same as another comment, should consider put this hard code into a global variable

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def prepare_try_build(self, revision):
+        url_try_builds = "https://archive.mozilla.org/pub/firefox/try-builds/"
+        req = urllib2.Request(url_try_builds)
+        response = urllib2.urlopen(req)
+        html_try_builds = response.read()
+
+        pattern = 'href="(.+' + message.revision + '\/)"'
+        m = re.search(pattern, html_try_builds)
+        url_revision_build = "https://archive.mozilla.org" + m.group(1) + "try-win64/firefox-55.0a1.en-US.win64.zip"
```
ShakoHo: You might need to add a TODO here, for supporting other platform in the future

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+        logger.debug("Download build from %s" % url_revision_build)
+        file_revision_build = urllib2.urlopen(url_revision_build)
```
ShakoHo: if user didn't submit the build for win64, you will raise an exception here..

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+            while True:
+                data = file_revision_build.read(4096)
+                if data:
+                    output.write(data)
+                else:
+                    break
+
+        with zipfile.ZipFile('firefox.zip', 'r') as zf:
+            zf.extractall()
+        shutil.rmtree(r'C:/Program Files/Mozilla Firefox')
```
ShakoHo: You might need to check the local PATH environment variable, some machine will set the firefox path to C:/Program Files (x86)/Mozilla Firefox...

agent/iskakalunan.py
```diff
@@ -0,0 +1,350 @@
...
+        except errors.HttpError, error:
+            logger.error("An error occurred: %s" % error)
+
+    def prepare_try_build(self, revision):
+        url_try_builds = "https://archive.mozilla.org/pub/firefox/try-builds/"
+        req = urllib2.Request(url_try_builds)
+        response = urllib2.urlopen(req)
+        html_try_builds = response.read()
+
+        pattern = 'href="(.+' + message.revision + '\/)"'
```
ShakoHo: not sure where did you get the variable message..?

PR #587, opened 2017-03-24
update .gitignore to filter noise

### LINKED ISSUE #586: Update .gitignore to avoid redundant information on git status

Conjuror:
fix #586 
update gitignore types

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,8 @@ docs/*
 *.flow
 *.bak
 *.txt
+! requirements.txt
+! requirements_mac.txt
 
 # 3rd party files
 thirdParty/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,17 @@
+# environment
 .DS_Store
 .idea
 .env/
 .env-*/
 env/
 env-*/
+.Hasal
+
+#local file
+agent_status/
+resource/
+upload/
+test.suite
 output
 snapshot
 result.json
@@ -15,6 +23,13 @@ _virtualenv
 HasalServer.log.*
 stage/*
 docs/*
+
+# File types to ignore
+*.zip
+*.diff
+*.flow
+*.bak
+*.txt
 
 # 3rd party files
 thirdParty/*


.gitignore
```diff
@@ -16,6 +24,14 @@ HasalServer.log.*
 stage/*
 docs/*
 
+# File types to ignore
+*.zip
+*.diff
+*.json
```
askeing: There are some JSON configuration files in Hasal, I think we shouldn't ignore JSON file.

Conjuror: I'll suggest we specific what we'd like to have in the repo. Currently, every thing is dump to the Hasal folder and it make the code hard to maintain.

Conjuror: For what I see now, svrConfig.json is the only json we should still monitor.

askeing: `settings` folder also has some json config files.
AFAIK, @ShakoHo will add more json config files to control the parameters in the future.

@ShakoHo , do you have any suggestions?

ShakoHo: The new Json settings will all store in configs folder...svrconfig.json will be removed after the new configs merged

.gitignore
```diff
@@ -1,9 +1,17 @@
...
 .env-*/
 env/
 env-*/
+.Hasal
+
+#local file
+agent_status/
+resource/
+upload/
+test.suite
```
ypwalter: why test.suite?

.gitignore
```diff
@@ -16,6 +24,14 @@ HasalServer.log.*
 stage/*
 docs/*
 
+# File types to ignore
+*.zip
+*.diff
+*.json
+*.flow
```
ypwalter: We do have some .flow files. We use them to run with mitmproxy and those are precached flows.
https://github.com/Mozilla-TWQA/Hasal/tree/dev/flows

.gitignore
```diff
@@ -16,6 +24,14 @@ HasalServer.log.*
...
+*.bak
+*.txt
```
ypwalter: We have 6 .txt files in root of Hasal directory.

.gitignore
Conjuror: @ypwalter test.suite is a local file for keeping the test suites which DuT is running. In case there will be diffs blocks code update. It's better to ignore test.suite. If we need to update default test.suite, it's fine to force update.

Conjuror:
@askeing I removed json from gitignore.
r? @askeing

.gitignore
```diff
@@ -16,6 +24,13 @@ HasalServer.log.*
 stage/*
 docs/*
 
+# File types to ignore
+*.zip
+*.diff
+*.flow
+*.bak
+*.txt
```
ShakoHo: can't ignore *.txt, that will also ignore requirement.txt etc.

Conjuror: good point, I'll exclude requirements.txt

PR #588, opened 2017-03-24
Auto mail robot for trigger jobs from e-mail

Conjuror:
#571 Propose a solution for auto mail robot

### DIFF
--- a/agent/iskakalunan.py
+++ b/agent/iskakalunan.py
@@ -6,7 +6,6 @@ import urllib2
 import os
 import sys
 import re
-import json
 import base64
 import subprocess
 import shutil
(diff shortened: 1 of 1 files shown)


agent/iskakalunan.py
```diff
@@ -0,0 +1,382 @@
...
+            logger.error('An error occurred: %s' % error)
+
+    def prepare_try_build(self, revision):
+        try:
+            url_try_builds = "https://archive.mozilla.org/pub/firefox/try-builds/"
+            req = urllib2.Request(url_try_builds)
+            response = urllib2.urlopen(req)
+            html_try_builds = response.read()
+
+            pattern = 'href="(.+' + message.revision + '\/)"'
```
askeing: `message.revision` should be `revision`

agent/iskakalunan.py
```diff
@@ -0,0 +1,382 @@
...
+            m = re.search(pattern, html_try_builds)
+            url_revision_build = "https://archive.mozilla.org" + m.group(1) + "try-win64/firefox-55.0a1.en-US.win64.zip"
```
askeing: it hardcode the download url to win64 build.
is it okay for other platform?

agent/iskakalunan.py
```diff
@@ -0,0 +1,382 @@
...
+            logger.debug("Download build from %s" % url_revision_build)
+            file_revision_build = urllib2.urlopen(url_revision_build)
```
askeing: Hasal's dependencies contain the `requests` package.
This package will help you download file easier.

```python
r = requests.get(url_revision_build, stream=True)
with open('firefox.zip', 'wb') as f:
    for chunk in r.iter_content(chunk_size= 4096):
        f.write(chunk)
```

agent/iskakalunan.py
Conjuror: Currently, the first release is targeted at Windows 64-bit. There should be a separate patch for migrating the code to support all platforms in the future.

PR #592, opened 2017-03-28
Support moving Terminal window for  Windows 7 and 10

askeing:
The offset and height of Terminal window are different on Window 7 and Windows 10.


askeing:
@ypwalter r?

ypwalter:
lgtm r+

PR #599, opened 2017-04-12
minor changes for backing up system path before installing

ypwalter:
Due to some previous experiences on PATH variable issues, this is to back up the variable first. 
I will do some more modification later to ensure PATH variable related issues are safe.

@ShakoHo r?

### DIFF
--- a/bootstrap.bat
+++ b/bootstrap.bat
@@ -1,7 +1,9 @@
+@REM This is mostly for windows 7. Recommend to run bootstrap.ps1 in Windows 10 instead.
 @REM Author: Walter Chen
 @REM Version: 0.1.1
 @REM 0.1.0 - Successfully running in virtual machines, appveyor, and real machines.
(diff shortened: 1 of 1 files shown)


bootstrap.bat
```diff
@@ -1,7 +1,9 @@
+@REM This is mostly for windows 7. Recommend to run bootstrap.ps1 in Windows 10 instead.
 @REM Author: Walter Chen
 @REM Version: 0.1.1
 @REM 0.1.0 - Successfully running in virtual machines, appveyor, and real machines.
 @REM 0.1.1 - Adding instruction and changed the command for detecting admin privileges.
+@REM 0.1.2 - Backing up the PATH variable for you in path.txt before running bootstrap
```
ShakoHo: how about using a better file name for this?  ex: bootstrap_backup_env_path.txt ?

ypwalter: Done.

ypwalter:
Thanks

PR #551, opened 2017-02-21
abstract BaseTest class

ypwalter:
@ShakoHo @askeing  r?

PR #600, opened 2017-04-12
rename sikuli_status to round_status

ypwalter:
sikuli_status was out-of-date.
The historic reason for this change is from #551 .
We no longer only support webdriver so we change the sikuli_status to round_status as a result.

Please do use round_status as returned status next time.

@ShakoHo @askeing  r?


ypwalter:
Thanks.

PR #605, opened 2017-04-13
update readme about virtualenv message (for master only)

ShakoHo:
@ypwalter r?
@Conjuror A specific fix for README.md on master..


ypwalter:
lgtm r+
However, should note somewhere else in commit that this is for current master branch only

PR #607, opened 2017-04-14
a tool to generate or remove post from facebook account 

ypwalter:
This is done by webdriver and should be immune from UI changes.
Please make sure to ran bootstrap for latest selenuim webdriver package and geckodriver for it.

In this case, we can do post of the same content for all tests.
@ShakoHo r?


ypwalter:
thanks

PR #608, opened 2017-04-14
Refactoring close browser

askeing:

### DIFF
--- a/lib/common/windowController.py
+++ b/lib/common/windowController.py
@@ -36,16 +36,16 @@ class WindowObject(object):
         # the identity only works on Linux now
         if self.window_type == "linux2":
             if current:
-                self.window_identity = self.wmctrl_get_current_window_id()
+                self.window_identity = self._get_current_window_id_linux()
             else:
-                self.window_identity = self.wmctrl_get_window_id()
+                self.window_identity = self._get_window_id_linux()
 
-    def wmctrl_get_current_window_id(self):
+    def _get_current_window_id_linux(self):
         cmd = "xprop -root 32x '\t$0' _NET_ACTIVE_WINDOW | cut -f 2"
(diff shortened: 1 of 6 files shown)


lib/common/windowController.py
```diff
@@ -173,11 +187,24 @@ def move_window_pos(self, pos_x=None, pos_y=None, window_width=None, window_heig
...
-            return self.appscript_move_window()
+            return self.move_window_mac()
         else:
             logger.warning('Doesn\'t support moving window [{}] on platform [{}].'.format(self.window_name_list, self.window_type))
             return False
+
+    def close_window(self):
+        # Try to move window
+        if self.window_type == 'linux2':
+            return self.close_window_linux()
```
ypwalter: 1. self.close_window_linux() -> self._close_window_linux(), and so does the other 2 function names.
2. close_window_(os) is interesting since there is a "close_window_win32". How about close_app_(os) instead?

lib/helper/desktopHelper.py
```diff
@@ -64,6 +63,31 @@ def launch_browser(browser_type, **kwargs):
     return browser_obj, profile_path
 
 
+def get_browser_name_list(browser_type):
+    if browser_type == Environment.DEFAULT_BROWSER_TYPE_FIREFOX:
+        if sys.platform == 'darwin':
+            return ['Firefox.app', 'FirefoxNightly.app']
+        else:
+            # This is to ensure all kinds of firefox we supported can be properly moved.
```
ypwalter: should we move this comment to line 67 or 68?

ypwalter:
r+ (but don't merge framework code on Friday)

lib/common/windowController.py
askeing: in my opinion, the 'window' is not 'Windows' or 'win32' :P
So, I prefer the way 1.

PR #610, opened 2017-04-14
support get platform dep settings from index config

ShakoHo:
extract search margin/threshold/fps to index config

@askeing @MikeLien r?


lib/generator/frameThroughputDctGenerator.py
```diff
@@ -206,8 +206,9 @@ def generate_result(self, input_data, input_global_result):
         compare_setting = {'default_fps': input_data['default_fps'], 'event_points': self.BROWSER_VISUAL_EVENT_POINTS,
```
MikeLien: `'default_fps': input_data['default_fps']` -> `'default_fps': input_data['index_config']['video-recording-fps']`

PR #613, opened 2017-04-17
Merge from dev

Conjuror:
plan to ship v1.0.0 for the new release.


Conjuror:
r? @ShakoHo 
per local test, everything works as expect. I believe we are good to ship this.

PR #614, opened 2017-04-17
[Travis] Disable Mac test-tidy on Max OSX

askeing:


askeing:
Disable `test-tidy` on Max OSX due to we already have it running on Linux.
It might speed up the CI response.

PR #596, opened 2017-04-05
Configuration setting refactor

ShakoHo:
@MikeLien @askeing r?

PR #622, opened 2017-04-18
extract case static variable from environment

MikeLien:
@ShakoHo, @askeing, @ypwalter r?


tests/regression/gdoc/test_chrome_gdoc_create_isolate_image_1.py
```diff
@@ -5,7 +5,7 @@ class TestSikuli(PerfBaseTest):
 
     def setUp(self):
         self.pre_run_script = "test_chrome_gdoc_read_blank_page"
-        self.pre_run_script_test_url_id = "TEST_TARGET_ID_BLANK_PAGE"
+        self.pre_run_script_test_url_id = self.global_config['gsuite']['gdoc-blank-page']
```
ypwalter: I think create_isolate_image_1 and create_image_1 are the same test case. So do create_isloate_table_1 and create_isolate_table_1. We should keep only one of them.

Also, pre_run_script was taken out in #596, and we probably should use the new `self.set_variable(test_target=self.global_config['gsuite']['gdoc-blank-page'])`

@MikeLien @ShakoHo @askeing, thoughts?

ypwalter:
I did a cross check for old and new hash matching. I think this pr is good. However, we might still want to solve the previously mentioned prescript issue.

tests/regression/gdoc/test_chrome_gdoc_create_isolate_image_1.py
ypwalter: According to @ShakoHo, it was used for debugging by thinker. It is for achieving something like the input latency measurement we have now. I suggest that you can remove these 6 files.

PR #626, opened 2017-04-19
update configuration from test cases

ypwalter:
@ShakoHo @askeing @MikeLien  r?


lib/baseTest.py
```diff
@@ -132,20 +137,29 @@ def remove_test_file(self):
...
+    def set_configs(self, config_variable_name, config_value, update=False):
         default_platform_dep_settings_key = "platform-dep-settings"
+        if default_platform_dep_settings_key in config_value:
+            if self.current_platform_name in config_value[default_platform_dep_settings_key]:
+                platform_dep_variables = copy.deepcopy(
+                    config_value[default_platform_dep_settings_key][self.current_platform_name])
+                config_value.update(platform_dep_variables)
+            config_value.pop(default_platform_dep_settings_key)
+
+        if update:
```
ShakoHo: we can use check variable exist in self-object to replace update variable

lib/baseTest.py
```diff
@@ -132,20 +137,29 @@ def remove_test_file(self):
         if hasattr(self, "test_url_id"):
             self.target_helper.delete_target(self.test_url_id)
 
-    def load_configs(self):
+    def set_configs(self, config_variable_name, config_value, update=False):
         default_platform_dep_settings_key = "platform-dep-settings"
+        if default_platform_dep_settings_key in config_value:
+            if self.current_platform_name in config_value[default_platform_dep_settings_key]:
```
ShakoHo: we can add platform version when getting value from json or update config

lib/baseTest.py
```diff
@@ -132,21 +136,39 @@ def remove_test_file(self):
...
         default_platform_dep_settings_key = "platform-dep-settings"
+        if default_platform_dep_settings_key in config_value:
+            # Load the index config's settings under "platform-dep-settings" base on platform
+            platform_dep_variables = self._platform_dep_settings_handler(config_value[default_platform_dep_settings_key])
+            config_value.update(platform_dep_variables)
+            config_value.pop(default_platform_dep_settings_key)
+
+        if hasattr(self, config_variable_name):
+            # getattr is a way to get variable by reference and doesn't need to be set back
+            getattr(self, config_variable_name).update(config_value)
```
askeing: I think we should check the accept config name.
And I will suggest modify it to two line for readable reason.

lib/baseTest.py
```diff
@@ -132,21 +136,39 @@ def remove_test_file(self):
...
+        if self.current_platform_name in config_value:
+            if self.current_platform_ver in config_value[self.current_platform_name]:
+                platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name][self.current_platform_ver])
+            else:
+                platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name][''])
+            config_value.update(platform_dep_variables)
+            return platform_dep_variables
+        return {}
+
+    def set_configs(self, config_variable_name, config_value):
```
askeing: I will suggest to add a class to record the accept config name
```python
class ConfigName():
    EXEC = 'exec_config'
    INDEX = 'index_config'
    # ...
```

And then check the config name before overwrite the original variable
```python
ACCEPT_CONFIG_LIST = [ConfigName.EXEC, ConfigName.INDEX, ...]
# ...
if config_variable_name not in ACCEPT_CONFIG_LIST:
    raise Exception('Do not accept overwrite the {config_name} : {config_value}'.format(config_name= config_variable_name, config_value=config_value))
```

Also, the cases can import `ConfigName` class to overwrite the settings easily
```python
self.set_configs(ConfigName.INDEX, {"platform-dep-settings": {"win32": {"": {"compare-threshold": 0.003}}}})
```

lib/baseTest.py
```diff
@@ -132,21 +137,48 @@ def remove_test_file(self):
         if hasattr(self, "test_url_id"):
             self.target_helper.delete_target(self.test_url_id)
 
-    def load_configs(self):
+    def platform_dep_settings_handler(self, config_value):
```
askeing: `foo_handler` is a noun.
In general, we use noun for class name, not method name.

lib/baseTest.py
```diff
@@ -132,21 +137,48 @@ def remove_test_file(self):
...
+        if self.current_platform_name in config_value:
+            if self.current_platform_ver in config_value[self.current_platform_name]:
+                platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name][self.current_platform_ver])
+            else:
+                platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name][''])
```
askeing: In my opinion, using `""` here as default value is too tricky. (Ref: https://github.com/Mozilla-TWQA/Hasal/pull/626/files#diff-daa66495dbc47f258f4c68bde193e7b3R16)

I'd like to use `"default"` for making code easier to understand.

lib/baseTest.py
ypwalter: It's a config setter though. The function is use to create or set a new config, and the config can be overwrite if the config created. How about ... `set_and_update_config()` then?

lib/baseTest.py
ypwalter: how about `extract_platform_dep_settings()`

lib/baseTest.py
ypwalter: agree. thanks.

lib/baseTest.py
askeing: 👍

lib/baseTest.py
askeing: I think maybe adding the clear comments is okay.

Any idea, @ShakoHo ?

ypwalter: Ok, adding more comments here and elsewhere in the code might be a good start.

ypwalter:
TODO for myself: need to squash commits into meaningful ones.

lib/baseTest.py
```diff
@@ -135,21 +140,49 @@ def remove_test_file(self):
...
+        if hasattr(self, config_variable_name):
+            # getattr is a way to get variable by reference and doesn't need to be set back
+            new_config_value = getattr(self, config_variable_name)
+            new_config_value.update(config_value)
+            setattr(self, config_variable_name, new_config_value)
+        else:
+            setattr(self, config_variable_name, config_value)
+
+    def load_configs(self):
+        config_fp_list = [ConfigName.EXEC_FP, ConfigName.INDEX_FP, ConfigName.GLOBAL_FP,
```
ShakoHo: you can move the ConfigName class to here, and store ConfigName into self object, so you test case doesn't have to import ConfigName

lib/baseTest.py
```diff
@@ -4,11 +4,13 @@
 import copy
 import time
 import json
+import platform
 import unittest
 import helper.desktopHelper as desktopHelper
 import lib.helper.videoHelper as videoHelper
 import lib.helper.targetHelper as targetHelper
 import helper.generatorHelper as generatorHelper
+from common.configName import ConfigName
```
askeing: I will suggest use Inner Class
ex:
```python
class Upper(object):

    class Inner(object):
        VAR = 'hello world'

    def run(self):
        print('Upper: {}'.format(self.Inner.VAR))


class Sub(Upper):
    def run(self):
        print('Sub: {}'.format(self.Inner.VAR))


myUpper = Upper()
myUpper.run()

mySub = Sub()
mySub.run()
```

And result
```bash
$ python temp/example.py
Upper: hello world
Sub: hello world
```

lib/baseTest.py
```diff
@@ -135,21 +140,49 @@ def remove_test_file(self):
...
+            setattr(self, config_variable_name, new_config_value)
+        else:
+            setattr(self, config_variable_name, config_value)
+
+    def load_configs(self):
+        config_fp_list = [ConfigName.EXEC_FP, ConfigName.INDEX_FP, ConfigName.GLOBAL_FP,
+                          ConfigName.FIREFOX_FP, ConfigName.ONLINE_FP]
+
         for config_env_name in config_fp_list:
             config_variable_name = "_".join(config_env_name.split("_")[:2]).lower()
```
askeing: Suggest using `rsplit`
```python
config_variable_name = config_env_name.rsplit('_', 1)[0].lower()
```

ex:
```python
>>> import time
>>> x = 'EXEC_CONFIG_FP'
>>> s = time.time(); "_".join(x.split("_")[:2]).lower(); e = time.time()
'exec_config'
>>> e-s
2.4080276489257812e-05
>>>
>>> s = time.time(); x.rsplit('_', 1)[0].lower(); e = time.time()
'exec_config'
>>> e-s
1.9073486328125e-05
```

lib/baseTest.py
ypwalter: Thanks. Will do.

lib/baseTest.py
ypwalter: I think that using python console with `time.time()` might not be that precise.

```
import timeit
print timeit.timeit('"_".join("EXEC_CONFIG_FP".split("_")[:2]).lower()', number=100000) 
print timeit.timeit('"_".join("EXEC_CONFIG_FP".rsplit("_", 1)[0].lower())', number=100000)
```

If you run the time measurement by timeit and get the accumulated time from 100000 times. The total time spend is less in `"_".join(x.split("_")[:2]).lower();`

> Ubuntu:
> (.env-python)hasal@hasal-BM6875-BM6675-BP6375:~/Hasal$ python test_timeit.py 
> 0.0785119533539
> 0.127026081085

> Windows:
> C:\Users\user\Hasal>python test_timeit.py
> 0.0833083045389
> 0.217570606103

I also tried to use time.time() and I ran it multiple times and sometimes changing the running order. I found that the result varies a lot....

ypwalter: Anyway, I think I would like to keep the old one.

askeing: `"_".join("EXEC_CONFIG_FP".rsplit("_", 1)[0].lower()` we shouldn't join again XD

askeing:
```python
>>> print timeit.timeit('"_".join("EXEC_CONFIG_FP".split("_")[:2]).lower()', number=100000)
0.0902090072632
>>> print timeit.timeit('"EXEC_CONFIG_FP".rsplit("_", 1)[0].lower()', number=100000)
0.0585010051727
>>>
```

askeing:
you should use `default` key in your cases.

ypwalter:
I forgot to pushed those files in. It was pushed into the repository around the same time you commented. Thanks a lot.

askeing:
lgtm

PR #631, opened 2017-04-24
adding facebook messenger wait functions and its related modification

ypwalter:
@MikeLien @ShakoHo @askeing 
messenger sometimes not loaded when  facebook title is loaded, this should fix it.


ypwalter:
@askeing  help~~

lib/sikuli/facebook.sikuli/facebook.py
```diff
@@ -52,6 +53,12 @@ def wait_for_loaded(self):
         wait(self.fb_logo)
         setAutoWaitTimeout(default_timeout)
 
+    def wait_for_messenger_loaded(self):
+        default_timeout = getAutoWaitTimeout()
+        setAutoWaitTimeout(10)
+        wait(self.messenger_header)
```
askeing: I am not sure what's the different between
```python
default_timeout = getAutoWaitTimeout()
setAutoWaitTimeout(10)
wait(self.messenger_header)
setAutoWaitTimeout(default_timeout)
```
and 
```python
wait(self.messenger_header, 10)
```

If they are the same, I will prefer use the simple one.

askeing: lgtm r+
we can send a new PR for wait timeout.

ypwalter: Thanks. You are right. According to the sikulix documentation, `wait(element, time)` is exactly what `setAutoWaitTimeout(time)` does.

Reference: http://doc.sikuli.org/region.html#Region.wait
> seconds – a number, which can have a fraction, as maximum waiting time in seconds. The internal granularity is milliseconds. If not specified, the auto wait timeout value set by Region.setAutoWaitTimeout() is used. Use the constant FOREVER to wait for an infinite time.

I will go ahead and make these old cleaner in next pr.

PR #632, opened 2017-04-24
update configuration from test cases

ypwalter:
#626 is having too many commits and I screwed up it by merging with dev branch frequently. I am creating this new branch so that the commit messages can be clean.
1. supporting changing of threshold in test case
2. having security exception if config is not in default list
3. making sys.platform in a new global variable
4. speeding up config_variable_name replacement

@askeing @ShakoHo @MikeLien 
Please review this one final time for me. Thanks.


ypwalter:
@ShakoHo @askeing Thanks for all the reviews and comments in #626

PR #635, opened 2017-04-24
OBS bootstrap for Windows 7 (batch file)

askeing:
- [x] skip if `%APPVEYOR%` is `True`
- [x] checking `C:\Program Files (x86)\obs-studio\bin\32bit\obs32.exe`, skip download and install if file exist
- [x] download `OBS-Studio-18.0.1-Full.zip`
- [x] unzip `OBS-Studio-18.0.1-Full.zip` to `C:\Program Files (x86)\obs-studio`
- [x] launch OBS for License Agreement

### DIFF
--- a/bootstrap.bat
+++ b/bootstrap.bat
@@ -196,6 +196,29 @@ del sikuli*.jar
 @echo on
 
 
+@REM Installing OBS
+
+IF "%APPVEYOR%"=="True" GOTO SkipOBS
+
+IF NOT EXIST "C:\Program Files (x86)\obs-studio\bin\32bit\obs32.exe" (
+    ECHO [INFO] Downloading OBS ...
+    pushd thirdParty
+    del OBS-Studio-18.0.1-Full.zip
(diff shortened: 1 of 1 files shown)


askeing:

bootstrap.bat
```diff
@@ -196,6 +196,29 @@ del sikuli*.jar
 @echo on
 
 
+@REM Installing OBS
+
+IF "%APPVEYOR%"=="True" GOTO SkipOBS
+
+IF NOT EXIST "C:\Program Files (x86)\obs-studio\bin\32bit\obs32.exe" (
```
ypwalter: What if they have 32-bit windows? Should check for both `Program Files` and `Program Files (x86)`

askeing: Hasal will always install into `C:\Program Files (x86)\obs-studio`.

Ref:
https://github.com/Mozilla-TWQA/Hasal/pull/635/files#diff-0214b751c5b767049c26b73be10348a1R209
https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/profiler/obsProfiler.py#L19-L39

ypwalter:
r+
I will make some changes after I finish my current works.

PR #638, opened 2017-04-24
Add chrome profile creator and fb open chat tab case

MikeLien:
@ShakoHo @askeing r?


runtest.py
```diff
@@ -298,7 +311,7 @@ def main():
     arguments = docopt(__doc__)
     run_test_obj = RunTest(exec_config=arguments['--exec-config'], firefox_config=arguments['--firefox-config'],
                            index_config=arguments['--index-config'], online_config=arguments['--online-config'],
-                           global_config=arguments['--global-config'])
+                           global_config=arguments['--global-config'], chrome_config=arguments['--chrome-config'])
```
askeing: It seems like we should add the `CHROME_CONFIG` into https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/baseTest.py#L162-L194

MikeLien: No, we don't need to add `CHROME_CONFIG` due to `FIREFOX_CONFIG` only be used by following method to check if there has video recording and fps validation. These parts should be extracted out from browser related config.

askeing: agree, we should extract recording config to `exec` or other config file.

lib/helper/chromeProfileCreator.py
```diff
@@ -0,0 +1,80 @@
...
+            return self._chrome_profile_path
+        else:
+            profile_path = self._create_chrome_profile()
+            self._download_cookies(cookies_settings=cookies_settings)
+            return profile_path
+
+    def _create_chrome_profile(self):
+        tmp_profile_dir = tempfile.mkdtemp(prefix='chromeprofile_')
+        logger.info('Creating Profile: {}'.format(tmp_profile_dir))
+        self.chrome_cmd = '{} --user-data-dir={}'.format(self.chrome_cmd, tmp_profile_dir)
```
ypwalter: `self.chrome_cmd = '{} --user-data-dir={}'.format(self.chrome_cmd, tmp_profile_dir)`

I am thinking if 
`self.chrome_cmd = '{} --no-startup-window --user-data-dir={}'.format(self.chrome_cmd, tmp_profile_dir)`
could work better.

By doing so, `close_browser(self.browser_type)` can be removed and the chrome process will terminated after created the profiler. I tried in Mac Mini and Windows, and it worked.

lib/helper/desktopHelper.py
```diff
@@ -38,8 +38,10 @@ def launch_browser(browser_type, **kwargs):
 
     if env.PROFILER_FLAG_CHROMETRACING in enabled_profiler_list:
```
ypwalter: what about:
```
if browser_type == env.DEFAULT_BROWSER_TYPE_CHROME:
    if env.PROFILER_FLAG_CHROMETRACING in enabled_profiler_list:
        profile_path = env.chrome_profile_path
        browser_obj = chrome_class(env.DEFAULT_BROWSER_HEIGHT, env.DEFAULT_BROWSER_WIDTH,
                                   tracing_path=env.chrome_tracing_file_fp,
                                   profile_path=profile_path)
    else:
        profile_path = env.chrome_profile_path
        browser_obj = chrome_class(env.DEFAULT_BROWSER_HEIGHT, env.DEFAULT_BROWSER_WIDTH, profile_path=profile_path)

elif browser_type == env.DEFAULT_BROWSER_TYPE_FIREFOX:
    if env.PROFILER_FLAG_FXTRACELOGGER in enabled_profiler_list:
        profile_path = env.firefox_profile_path
        browser_obj = firefox_class(env.DEFAULT_BROWSER_HEIGHT, env.DEFAULT_BROWSER_WIDTH, tracelogger=True,
                                    profile_path=profile_path)
    else:
        profile_path = env.firefox_profile_path
        browser_obj = firefox_class(env.DEFAULT_BROWSER_HEIGHT, env.DEFAULT_BROWSER_WIDTH,
                                    profile_path=profile_path)
```

This might reduce some duplicated code.

askeing: Sending a new PR for this change should be better than adding to this PR.

ypwalter: Agreed. I think @MikeLien can still choose if he wants to do this in his PR, or I will do another PR after this PR is in.

MikeLien: I'll postpone this change to next PR to separate the purpose of code change.

PR #639, opened 2017-04-25
Remove Sikulix setAutoWaitTimeout() in common library

ypwalter:
@askeing @MikeLien @ShakoHo r?

According to the sikulix documentation, `wait(element, time)` is exactly what `setAutoWaitTimeout(time);wait(element);` does.

Reference: http://doc.sikuli.org/region.html#Region.wait
> seconds – a number, which can have a fraction, as maximum waiting time in seconds. The internal granularity is milliseconds. If not specified, the auto wait timeout value set by Region.setAutoWaitTimeout() is used. Use the constant FOREVER to wait for an infinite time.

However, there are still more `setAutoWaitTimeout()` in pilot or regression test cases. I think we should consider a more elegant way to handle them when we create new cases and when we have to modify the test cases. The code should be like what I did in this pr.


ypwalter:
Thanks.

Issue #633, opened 2017-04-24
Refactoring Windows 7 and 10 bootstrap


askeing:
This is note issue for `bootstrap.bat` refactoring

- [x] Win7 FFmpeg: Check the binary before download and install. [Code Ref](https://github.com/Mozilla-TWQA/Hasal/blob/dev/bootstrap.bat#L163-L178) (@askeing #636) 
- [x] Win10 PATH overcoming the 1024 character limit (@ypwalter #641 #633)
- [x] Win7 PATH overcoming the 1024 character limit (@ypwalter #740)
- [x] Win7 Win10: Check OBS before install it (@askeing #652)
- [ ] Win7 handling of dup in PATH

PR #652, opened 2017-04-26
Check OBS binary before install it

askeing:
ref #633

### DIFF
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -90,16 +90,21 @@ If ($lastexitcode -notmatch 0) {
 
 
 # Install OBS Studio
-"[INFO] Downloading OBS ..."
-pushd thirdParty
-If (Test-Path OBS-Studio-18.0.1-Full.zip) {
-    Remove-Item OBS-Studio-18.0.1-Full.zip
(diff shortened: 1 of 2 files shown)


askeing:
@ypwalter @MikeLien r?

bootstrap.ps1
```diff
@@ -90,16 +90,21 @@ If ($lastexitcode -notmatch 0) {
...
-"[INFO] Downloading OBS ..."
-pushd thirdParty
-If (Test-Path OBS-Studio-18.0.1-Full.zip) {
-    Remove-Item OBS-Studio-18.0.1-Full.zip
+If (Test-Path "C:\Program Files (x86)\obs-studio\bin\32bit\obs32.exe") {
+    "[INFO] OBS had been installed."
+} Else {
+    "[INFO] Downloading OBS ..."
+    pushd thirdParty
+    If (Test-Path "OBS-Studio-18.0.1-Full.zip") {
```
ypwalter: Do you also want to remove it after installation, too?

ypwalter: I should also make a Check-Remove function since we are using this Test-Path and Remove-Item too frequently. It could be a function for further use.

askeing: I don't think we have to remove it, the folder `thirdparty` is listed in `.gitignore`.

PR #654, opened 2017-04-27
support running without client certificate

ShakoHo:
@MikeLien @askeing @ypwalter r?
- add enable the option for firefox/chrome profile creation
- add checking with client certificate within PyDriveUtil.py
- move test_target to create test target function

### DIFF
--- a/configs/chrome/default.json
+++ b/configs/chrome/default.json
@@ -1,4 +1,5 @@
 {
+  "enable_create_new_profile": true,
   "cookies": {
     "folder"              : "0BwkEhia_D6l_d2JoSGNhS3VLblE",
     "Cookies"             : "Cookies_gsuite_fb_linda",
--- a/lib/common/pyDriveUtil.py
+++ b/lib/common/pyDriveUtil.py
@@ -1,3 +1,4 @@
+import os
 import argparse
 from pydrive.auth import GoogleAuth
 from pydrive.drive import GoogleDrive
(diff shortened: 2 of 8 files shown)


configs/chrome/default.json
```diff
@@ -1,4 +1,5 @@
 {
+  "enable_create_new_profile": true,
```
ypwalter: I think this should be applied to all config json files under `configs/chrome/` and `configs/firefox/`?

lib/common/pyDriveUtil.py
```diff
@@ -12,8 +13,12 @@ class PyDriveUtil(object):
...
         if settings is None:
             settings = {"settings_file": None, "local_cred_file": "mycreds.txt"}
-        self.gauth = self.get_gauth(settings)
-        self.drive = GoogleDrive(self.gauth)
+        if os.path.exists(settings['local_cred_file']):
+            self.gauth = self.get_gauth(settings)
+            self.drive = GoogleDrive(self.gauth)
+        else:
+            raise Exception(
+                "Your current working dir didn't inculde the client certificate file [%s], please make sure the firefox/chrome profile creation is turned off, online mode is disabled and not running the test cases of GSuite and FB!")
```
ypwalter: Your current working dir `doesn't` `include` the client certificate file [%s]`.` Please make sure firefox/chrome profile creation is turned off, online mode is disabled`,` and `GSuite and FB cases are not test targets!`

lib/common/pyDriveUtil.py
```diff
@@ -12,8 +13,12 @@ class PyDriveUtil(object):
     def __init__(self, settings=None):
         if settings is None:
             settings = {"settings_file": None, "local_cred_file": "mycreds.txt"}
-        self.gauth = self.get_gauth(settings)
-        self.drive = GoogleDrive(self.gauth)
+        if os.path.exists(settings['local_cred_file']):
+            self.gauth = self.get_gauth(settings)
```
ypwalter: `self.gauth` is not used elsewhere, I think you can use gauth instead like the refactor you made in:
https://github.com/Mozilla-TWQA/Hasal/pull/654/files#diff-9a13e7681f813c5a68e0ed49a6c4935fR89

ShakoHo:
fix #654

PR #658, opened 2017-05-02
iskakalunan for local CI test

Conjuror:


agent/agent.py
```diff
@@ -106,8 +106,8 @@ def load(self, fp):
 
         else:  # Create new
             logger.info("Create new runner [%s]" % fp)
-            module_path = "hasalTask"
-            object_name = "HasalTask"
+            module_path = "iskakalunan"
+            object_name = "Iskakalunan"
```
ShakoHo: It changes to hard code the module path to iskkaklunan, I will suggest we can put a parameter to control this..

Conjuror: @ShakoHo I'll suggest we create a separate issue to track this one instead of having one PR solves all problems.

Conjuror: Issue #682 is introduced for solving the change.

agent/githubHook.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class githubHook(object):
+    def __init__(self):
+        user, password = self._get_credentials()
+        g = github3.login(user, password)
+        self.hasal = g.repository('Mozilla-TWQA', 'Hasal')
+
+    def _get_credentials(self):
+        github_credential = 'github.json'
```
askeing: Please adding the comment for describing the structure of github.json.

By the way, how about using the authorization token to replace the username/pwd?

agent/agent.py
ShakoHo: I can have another fix to solve this issue, since I want to modify agent.py this week..

agent/githubHook.py
Conjuror: The reason works for me for keeping perfherder jobs work. I'll modify this back.

For credential parts, I agree it would be better if we can use authorization token. Do you think this is a must-have for draft? @askeing

askeing: IMO, adding the comment for describing the `github.json` for draft is okay now.

Conjuror: Issue #682 is introduced for using json format file to decide runner on executing.
Issue #683 is introduced for using authorization token.

Conjuror:
fix #683

PR #667, opened 2017-05-03
Adding Sikuli general WebApp class for general actions

askeing:
- Adding `common.WebApp` class
- Refactoring Sikuli's `common.General` class


askeing:
@MikeLien @ypwalter 
It might help the implement of all Web Apps general classes, r?

PR #668, opened 2017-05-04
modified threshold and fb case library for win10

ypwalter:
@ShakoHo @MikeLien r?


lib/sikuli/facebook.sikuli/facebook.py
```diff
@@ -43,7 +43,7 @@ def __init__(self):
         self.share_menu = Pattern("pics/facebook_share_menu.png").similar(0.70)
         self.save_button = Pattern("pics/facebook_save_button.png").similar(0.70)
         self.club_post_menu_delete = Pattern("pics/facebook_club_post_menu_delete.png").similar(0.70)
-        self.messenger_header = Pattern("pics/facebook_messenger_header.png").similar(0.70)
+        self.messenger_header = Pattern("pics/facebook_messenger_header.png").similar(0.40)
```
ShakoHo: if would better using Askeing's lib to loop component image instead of lower the similarity.

ypwalter: Got it. Will do it with other images ( for gsearch suite, too )

ypwalter:
thanks

PR #673, opened 2017-05-04
Creating cases' result folder before generating case result

askeing:
I found the Hasal will raise exception when we had some original cases result data in `result.json`, and the current `suite.txt` does not contain that cases.

ex:
1. The `result.json` has result of `A`, `B`.
2. Running `suite.txt`, which contains `C`.
3. The `result.json` has result of `A`, `B`, `C`.
4. Generating ipynb and json files for each cases, however current `result` folder doesn't have `A` and `B` sub-folder. Raise exception.


askeing:
@ShakoHo r?

PR #674, opened 2017-05-04
Support getting current Username, and fixing the username issue of obsProfiler

askeing:
We found the `USERNAME` of each Windows machines might be different from default name `user`.

So I try to implement the static methods for getting username, AppData folder, user home folder.
And then modifying the `obsProfile`'s paths.


askeing:
@ShakoHo r?

PR #675, opened 2017-05-04
Creating 'BasicWindow' section if it not exists, or set() will raise NoSectionError

askeing:
Sometimes `obsProfiler` will raise `NoSectionError` when it try to execute `_modify_obs_global_settings()`.

The reason is [ConfigParser.set()](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.set) does not handle the case of section not exists.
```
RawConfigParser.set(section, option, value)
If the given section exists, set the given option to the specified value; otherwise raise NoSectionError.
```

So, adding the code for check the section and create it if need.

### DIFF
--- a/lib/profiler/obsProfiler.py
+++ b/lib/profiler/obsProfiler.py
@@ -96,6 +96,10 @@ class ObsProfiler(BaseProfiler):
             profile_config = self._get_obs_config_ini(basic_ini_src_fp)
             for section in input_settings:
                 for option in input_settings[section]:
+                    # create section if need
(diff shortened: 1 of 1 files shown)


askeing:
@ShakoHo @ypwalter r?

ypwalter:
I am not sure about this. I will try to do further testing tomorrow.
The way Shako shared me is to change some video output settings so I am not sure if this is the solution. I might need to clean AppData and some machine(maybe the one Al is currently using for testing this patch out)

Thanks.

PR #676, opened 2017-05-04
Fix logic error for threshold customization

ypwalter:
if running in win x without win y in win 8, customized threshold would throw exception. it should return {}.

@ShakoHo @askeing  r?

### DIFF
--- a/lib/baseTest.py
+++ b/lib/baseTest.py
@@ -151,19 +151,18 @@ class BaseTest(unittest.TestCase):
         if hasattr(self, "test_url_id"):
             self.target_helper.delete_target(self.test_url_id)
 
+    # be careful to have "default" value for each and every platform in conf file, or it would raise exception.
     def extract_platform_dep_settings(self, config_value):
+        platform_dep_variables = {}
+        # get current platform value or default value of it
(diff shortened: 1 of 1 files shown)


lib/baseTest.py
```diff
@@ -153,11 +153,15 @@ def remove_test_file(self):
...
+            # get current platform value or try to see if there is an default value
             if self.current_platform_ver in config_value[self.current_platform_name]:
                 platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name][self.current_platform_ver])
-            else:
+            elif 'default' in config_value[self.current_platform_name]:
                 platform_dep_variables = copy.deepcopy(config_value[self.current_platform_name]['default'])
-            config_value.update(platform_dep_variables)
+            # if not value doesn't match customized platform in test case return {}
+            else:
+                return {}
```
askeing: We might just remove `else` part, and set up the `platform_dep_variables` to default `{}`.
Finally, return `platform_dep_variables`.

ypwalter: thanks!

ypwalter:
thanks

PR #677, opened 2017-05-05
Modifying the GMail cases' wait time

askeing:
Wait more than one second before Input Latency Action.

### DIFF
--- a/lib/sikuli/common.sikuli/common.py
+++ b/lib/sikuli/common.sikuli/common.py
@@ -125,11 +125,25 @@ class WebApp(object):
         @param component: The waiting component
         @param similarity: The similarity of component. Default: 0.70.
         """
+        is_exists = False
         for counter in range(50):
+            if is_exists:
+                break
             for pic, _, _ in component:
                 if exists(Pattern(pic).similar(similarity), 0.1):
                     break
         wait(Pattern(pic).similar(0.70), 1)
+
+    def il_type(self, message, width, height, wait_component=None):
+        # wait component exists
+        if not wait_component:
+            self._wait_for_loaded(wait_component)
+
+        # Screenshot and get time for Input Latency
+        action_name = '[log]  TYPE "{}"'.format(message)
+        screenshot, current_time = self._screenshot_and_time(width=width, height=height, action_name=action_name)
+        type(message)
+        return screenshot, current_time
 
     def _click(self, action_name, component, similarity=0.70):
         """
--- a/tests/regression/gmail/test_firefox_gmail_ail_type_in_reply_field.sikuli/test_firefox_gmail_ail_type_in_reply_field.py
+++ b/tests/regression/gmail/test_firefox_gmail_ail_type_in_reply_field.sikuli/test_firefox_gmail_ail_type_in_reply_field.py
@@ -39,13 +39,13 @@ my_browser.enterLink(INPUT_TEST_TARGET)
 app.wait_for_loaded()
 
 # Wait for stable
-sleep(5)
+sleep(2)
 
 # PRE ACTIONS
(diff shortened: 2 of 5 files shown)


askeing:

tests/regression/gmail/test_firefox_gmail_ail_type_in_reply_field.sikuli/test_firefox_gmail_ail_type_in_reply_field.py
```diff
@@ -39,13 +39,13 @@
 app.wait_for_loaded()
 
 # Wait for stable
-sleep(5)
+sleep(2)
 
 # PRE ACTIONS
 app.click_first_mail()
```
ypwalter: I suggest that you can wait for reply button to show up for `app.click_first_mail()` and wait for submit button to show up for `app.click_reply_btn()`. Sometimes, the could still not fully loaded after a short wait, and the system will do the ` com.system_print('[log]  TYPE "a"')` right away. It could cause error and I did met it in Facebook cases.

askeing: The `WebApp._click()` already wait for component exists.
So, in this case, it will
1. wait 'compose btn' <= `app.wait_for_loaded()`
2. wait 'more btn', and then click first mail <= `app.click_first_mail()`
3. wait 'reply btn', and then click reply btn <= `app.click_reply_btn()`

https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/sikuli/common.sikuli/common.py#L145

askeing: Well, I got your point now.

The real problem point is
```python
app.click_reply_btn()
sleep(3)
 
# Record T1, and capture the snapshot image
# Input Latency Action
t1 = time.time()
screenshot = capture(0, 0, capture_width, capture_height)
 
com.system_print('[log]  TYPE "a"')
type('a')
```

We should wait the `send btn` exist then type a.
It seems like I can add a `_il_type(message, wait_component=None)` method for typing cases.

👍

askeing:
Adding `WebApp.il_type()` so that we can wait for `SendBtn` before type the message into `Reply Field`.

```python
screenshot, t1 = app.il_type('a', capture_width, capture_height, wait_component=app.GMAIL_SEND)
```

lib/sikuli/common.sikuli/common.py
```diff
@@ -125,12 +125,26 @@ def _wait_for_loaded(self, component, similarity=0.70):
         @param component: The waiting component
         @param similarity: The similarity of component. Default: 0.70.
         """
+        is_exists = False
         for counter in range(50):
```
ypwalter: 5 second might be okay for gmail. However, in FB, I encountered a case that it was delayed for around 10 seconds. Can you try to  make the function `_wait_for_loaded(self, component, time=5, similarity=0.7)` & `il_type(self, message, width, height, wait_component=None, time=5)` and `range(time*10)` in the for loop below please? In that case, I can reuse it in FB cases and other cases.

Thank you.

askeing: That's great idea.

However, there is a bad news that, the wait time `N` will take more than `N` seconds.
The reason is each image will wait `0.1` sec, see following code:
```python
             for pic, _, _ in component:
                  if exists(Pattern(pic).similar(similarity), 0.1):
                      break
```
So, when we have more than 1 image in the component, it will take more than wait time `N`.
I'll plan to do the overall code change in #679.

PR #678, opened 2017-05-05
Adding Terminal Crop Region for comparing images

askeing:
- Adding terminalHelper for handling Terminal location.
- Adding imageUtil.CropRegion for Crop Region's naming.
- Adding visualmetricsWrapper.find_terminal_view for comparing.
- Modifying inputLatencyAnimationDctGenerator for finding "start" by Terminal Region

Few weeks ago, @ShakoHo, @MikeLien and I have a discussion on comparing images for finding events, e.g. "start" and "end".
We found the "start" event of Input Latency cases by comparing the full screenshot (original image).
However, the IL cases will print the LOG on the Terminal when triggering the action.
We only want to compare the Terminal Region to find the "start" event for preciseness.
That's why I try to crop the Terminal Region from Original images.

### DIFF
--- a/lib/common/imageUtil.py
+++ b/lib/common/imageUtil.py
@@ -14,6 +14,37 @@ from logConfig import get_logger
 logger = get_logger(__name__)
 
 
+class CropRegion(object):
+    """
+    Here's the Crop Region Fig
+
+    |----------------------------------|  ---             ---
+    |        |                         |   ^               ^
+    | Tab  X |                         |   V <= TAB_VIEW   |
+    |----------------------------------|  ---              |
+    |                                  |   ^               |
+    |                                  |   |               |
+    |                                  |   |               |
+    |            Web Page              |   | <= VIEWPORT   | <= BROWSER
+    |                                  |   |               |
+    |                                  |   |               |
+    |                                  |   v               v
+    |----------------------------------|  ---             ---
+
+    |----------------------------------|  ---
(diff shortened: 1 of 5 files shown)


askeing:
@ShakoHo @MikeLien r?

lib/common/imageUtil.py
```diff
@@ -14,6 +14,37 @@
 logger = get_logger(__name__)
 
 
+class CropRegion(object):
+    """
```
ypwalter: 👍

lib/common/visualmetricsWrapper.py
```diff
@@ -49,12 +50,43 @@ def colors_are_similar(a, b, threshold=30):
...
+        base_y = int(viewport.get('y', 0))
+        base_width = int(viewport.get('width', im_width))
+        base_height = int(viewport.get('height', 0))
+
+        # get Terminal location
+        terminal_location = get_terminal_location(base_x, base_y, base_width, base_height)
+
+        # Adjust the height (do not over the image height)
+        terminal_y = terminal_location.get('y')
+        bottom = min(im_height, terminal_y + terminal_location.get('height') + 25)
```
ShakoHo: Could this shift value 25 might be a platform dependency value?

askeing: This shift offset is difference between different base, `browser` at runtime and `viewport` at screenshot.

PR #680, opened 2017-05-06
Supporting the waiting timeout of Sikuli WebApp

### LINKED ISSUE #679: Support customizing the timeout of `WebApp`
As @ypwalter said:
```
5 second might be okay for gmail. However, in FB, I encountered a case that it was delayed for around 10 seconds. Can you try to make the function _wait_for_loaded(self, component, time=5, similarity=0.7) & il_type(self, message, width, height, wait_component=None, time=5) and range(time*10) in the for loop below please? In that case, I can reuse it in FB cases and other cases.

Thank you.
```
The `WebApp` of `lib/sikuli/common.sikuli/common.py` should customize the timeout.

askeing:
Fix #679

From the begining, I plan to write two decorators, `WaitBeforeAction` and `WaitAfterAction`[1], for customizing the waiting timeout of each WebApp's action methods.
However, it seems like we only have to implement new parameters, `timeout` and `wait_component`, to fit the requirement.
Finially, I add the new parameters, and re-design the wait loop base on the pattern amount of components.

P.S
[1] Usage example:
```python
@WaitBeforeAction(component=GMail.FooBtn, timeout=20, similarity=0.85)
def click_bar_btn(self):
    do_somthing()
    ...
```

### DIFF
--- a/lib/sikuli/common.sikuli/common.py
+++ b/lib/sikuli/common.sikuli/common.py
@@ -119,25 +119,42 @@ class WebApp(object):
             self.control = Key.CTRL
             self.alt = Key.ALT
 
-    def _wait_for_loaded(self, component, similarity=0.70):
+    @staticmethod
+    def _get_loop_seconds(object_amount, total_second):
         """
-        Wait for component loaded, max 5 sec
-        @param component: The waiting component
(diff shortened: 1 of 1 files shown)


askeing:

PR #681, opened 2017-05-08
Resize browser to expected viewport

MikeLien:

### DIFF
--- a/lib/baseTest.py
+++ b/lib/baseTest.py
@@ -16,6 +16,7 @@ from common.logConfig import get_logger
 from common.windowController import WindowObject
 from common.commonUtil import CommonUtil
 from common.commonUtil import StatusRecorder
+from common.visualmetricsWrapper import find_image_viewport
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


lib/baseTest.py
```diff
@@ -195,6 +208,34 @@ def load_configs(self):
...
+            'video-recording-pos-x': self.env.DEFAULT_VIDEO_RECORDING_POS_X,
+            'video-recording-pos-y': self.env.DEFAULT_VIDEO_RECORDING_POS_Y,
+            'video-recording-width': self.env.DEFAULT_VIDEO_RECORDING_WIDTH,
+            'video-recording-height': self.env.DEFAULT_VIDEO_RECORDING_HEIGHT
+        }
+        self.set_configs(self.config_name.EXEC, screen_settings)
+
+    def get_new_recording_size(self):
+        recording_width = min(self.exec_config['browser-width'] + 100, 1920)
+        if platform.system().lower() == "darwin":
```
ypwalter: There is a global variable `self.current_platform_name` for platform name.

lib/common/windowController.py
```diff
@@ -196,6 +239,18 @@ def move_window_pos(self, pos_x=None, pos_y=None, window_width=None, window_heig
             logger.warning('Doesn\'t support moving window [{}] on platform [{}].'.format(self.window_name_list, self.window_type))
             return False
 
+    def get_window_rect(self):
```
askeing: IMO, we should initialize `self.rect` at `__init__()`.

PR #685, opened 2017-05-10
Customizing the InputLatencyCalcutionUtil for 0 frame issue when start and end at the same time, and modifying the event comparing result list data structure

askeing:
- Customizing the InputLatencyCalcutionUtil for 0 frame issue when `start` and `end` at the same time.
- Modifying the return data structure of event comparing result list from `[{'start': 'foo/bar/img.bmp', 'time_seq': 9527}, {...}]` to `[{'event': 'start', 'file': 'foo/bar/img.bmp', 'time_seq': 9527}, {...}]`.
- Refactoring the `runtime` and `frameThroughput` generators for new event comparing result list data structure. Also modifying the `get_frame_throughput()` for `frameThroughput` generator.
- Fixing the inconsistent issue of `search-range` and `converted images length` when CV2 converter failed.
- Resetting the current image-index when we start to compare the next event's images.
- Modifying the `tearDown()` for supporting the 2nd snapshot with multiple recorders.

### DIFF
--- a/lib/common/commonUtil.py
+++ b/lib/common/commonUtil.py
@@ -65,24 +65,48 @@ class StatusRecorder(object):
 
 class CalculationUtil(object):
 
+    EVENT_START = 'start'
+    EVENT_END = 'end'
+
+    @staticmethod
+    def get_event_data_in_result_list(event_result_list, event_name):
+        """
+        Return the event object which
+        @param event_result_list: the running_time_result after do comparison.
+            ex:
+            [
+                {'event': 'start', 'file': 'foo/bar/9487.bmp', 'time_seq': 5487.9487},
+                {'event': 'end', 'file': 'foo/bar/9527.bmp', 'time_seq': 5566.5566}, ...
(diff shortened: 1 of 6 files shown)


askeing:
@ShakoHo @MikeLien r?

lib/common/commonUtil.py
```diff
@@ -65,9 +65,52 @@ def record_status(self, case_name, status, value):
...
     @staticmethod
     def runtime_calculation_event_point_base(input_running_time_result):
         run_time = -1
+        event_time_dict = dict()
+
+        start_event = CalculationUtil.get_event_data_in_result_list(input_running_time_result,
+                                                                    CalculationUtil.EVENT_START)
+        end_event = CalculationUtil.get_event_data_in_result_list(input_running_time_result,
+                                                                  CalculationUtil.EVENT_END)
+        if start_event and end_event:
```
MikeLien: should be `or`

PR #687, opened 2017-05-10
Moving the running time calculation method from `commonUtil.CalculationUtil` to `basGenerator`

askeing:
After having a discussion with @ShakoHo and @MikeLien, we think it will be more reasonable if we move the running time calculation method from CalculationUtil to baseGenerator.
When sub-generators have to customize their own calculating methods, then they can override the "calculate_runtime_base_on_event" to fit the requirements.

### DIFF
--- a/lib/common/commonUtil.py
+++ b/lib/common/commonUtil.py
@@ -65,50 +65,6 @@ class StatusRecorder(object):
 
 class CalculationUtil(object):
 
-    EVENT_START = 'start'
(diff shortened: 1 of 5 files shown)


askeing:
@ShakoHo @MikeLien 
r?

askeing:
After having discussion with @ShakoHo , we all agree that we only have to move the running time calculation method from `commonUtil` to `baseGenerator`.

PR #690, opened 2017-05-11
tiny fix for issues

Conjuror:


agent/githubHook.py
```diff
@@ -59,4 +59,4 @@ def checkout_pr(pr):
                 f.write(auth.token + '\r\n')
                 f.write(str(auth.id))
         except Exception as e:
-            print e
+            print e.response.json()
```
askeing: The original `Exception` does not contain the `response` object.
It will raise another Exception when `e` is not `GitHubError`, for example, file-related exceptions.

Conjuror:
roll back to original version, since it's not a normal entry for githubHook.
r? @askeing

askeing:
lgtm, r+

PR #691, opened 2017-05-11
Adding the logging message for recording Comparing Threshold

askeing:
It will help us to debug or finding the precise comparing threshold.


askeing:
@ShakoHo @MikeLien r?

PR #693, opened 2017-05-12
Modifying the wait_for_loaded method

askeing:
@MikeLien found the `exista(pattern)` will take almost 0.5 sec (5xx ms) for checking the pattern.
It seems like we should get the more precise loop counter for waiting components.


askeing:
@MikeLien r?

PR #697, opened 2017-05-15
Adding the logging when current image can not be converted to DCT object

askeing:
Currently we will get the confusing error message `None type has no attribute shape` at `compare_two_images()` when DCT object is None.
We should check DCT objects before comparing them.
And also adding more clear docstring for comments.


askeing:
@MikeLien @ShakoHo r?

PR #699, opened 2017-05-16
Fix the output video issue of modifying the result data structure

askeing:
We modified the event comparing result list data structure, but did not change the `output_runtime_result_video` which will refer to the file path of `START` and `END` events.
So, we can not get the running time result video (which only contains `START` to `END` event).
After this fix, this feature will back.

Also, adding the docstring and comments.

Ref: #685


askeing:
@MikeLien r?

PR #701, opened 2017-05-16
Retry cases when we found the `Start` after `End` event

askeing:
Sometimes the `Start` will be found after `End` event due to the comparison threshold, we should catch this error and retry the case.


askeing:
@ShakoHo @MikeLien r?

PR #703, opened 2017-05-16
Return the empty object if there is no matched event in result list

askeing:
Sometimes the event object have to get the attribute from event object:
```python
        # get start point and end point from input data
        start_event = self.get_event_data_in_result_list(running_time_result,
                                                         self.EVENT_START)
        end_event = self.get_event_data_in_result_list(running_time_result,
                                                       self.EVENT_END)
        start_fp = start_event.get('file', None)
        end_fp = end_event.get('file', None)
```


askeing:
@ShakoHo r?

PR #708, opened 2017-05-18
fix missing self container and modify default branch to dev for testi…

Conjuror:
…ng purpose

### DIFF
--- a/agent/iskakalunan.py
+++ b/agent/iskakalunan.py
@@ -66,7 +66,7 @@ class Iskakalunan(object):
             subprocess.call(['python', 'runtest.py', '-h', '>', 'result'])
             return
 
-        subprocess.call(['git', 'checkout', 'master'])
+        subprocess.call(['git', 'checkout', 'dev'])
         # is there any new Mail request
         messages = self.mail_handler.get_requests()
         if messages:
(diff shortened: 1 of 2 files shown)


agent/iskakalunan.py
```diff
@@ -66,7 +66,7 @@ def run(self):
             subprocess.call(['python', 'runtest.py', '-h', '>', 'result'])
             return
 
-        subprocess.call(['git', 'checkout', 'master'])
+        subprocess.call(['git', 'checkout', 'dev'])
```
ypwalter: I think we should probably have some configuration file so that it can be more flexible on choosing the branch to be checked out?

Conjuror: by default, this should be master as always. The purpose to move to master was for local testing. Per Shako's comment, we were expecting to push local builds/try builds to e-mail account and be tested locally. If that's the case, it should run on master.

However, we are slow in updating master. That's the reason I propose to move current working branch to dev. Supposedly, we should move back to master in the future.

ypwalter: I am thinking of reading a configuration file so that we don't need to commit code to github for switching branches. In that way, we can even switch to do different branches easily. However, it should be fine for now.

ShakoHo: @Conjuror  if you want this parameter able to read from config, you can just simply add the key value pair in the config file, and you can get that variable from kwargs in __init__ function.

PR #710, opened 2017-05-19
support clean unnecessary image files

MikeLien:

### DIFF
--- a/lib/generator/frameThroughputDctGenerator.py
+++ b/lib/generator/frameThroughputDctGenerator.py
@@ -268,9 +268,4 @@ class FrameThroughputDctGenerator(BaseGenerator):
             elapsed_time = current_time - start_time
             logger.debug("Generate Video Elapsed: [%s]" % elapsed_time)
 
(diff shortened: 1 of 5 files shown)


lib/generator/frameThroughputDctGenerator.py
```diff
@@ -267,3 +267,10 @@ def output_case_result(self, suite_upload_dp):
             current_time = time.time()
             elapsed_time = current_time - start_time
             logger.debug("Generate Video Elapsed: [%s]" % elapsed_time)
+
```
ypwalter: why don't we just put 
```
if self.exec_config['clean-unnecessary-images']:
    start_time = time.time()
    self.clean_output_images(self.compare_result['running_time_result'], self.env.img_output_dp)
    current_time = time.time()
    elapsed_time = current_time - start_time
    logger.debug("Clean Output Images Elapsed: [%s]" % elapsed_time)
```
inside clean_output_images()?
After doing so, we can just call clean_output_images here instead.

PR #712, opened 2017-05-22
output case exec time and fps to result files

ShakoHo:


ShakoHo:

lib/common/windowController.py
```diff
@@ -162,7 +162,7 @@ def _get_window_rect_darwin(self):
                             if application_obj.isrunning():
                                 # move window position
                                 # Reference: https://www.macosxautomation.com/applescript/firsttutorial/11.html
-                                rect = application_obj.windows.bounds.get()
+                                rect = application_obj.windows.bounds.get()[0]
```
askeing: How about checking the type (list) and length (>0) of `application_obj.windows.bounds.get()` before getting the values?

lib/common/commonUtil.py
```diff
@@ -4,27 +4,26 @@
...
-from datetime import datetime
+import time
 from environment import Environment
 from logConfig import get_logger
-
+from datetime import datetime
 logger = get_logger(__name__)
 
 
 class StatusRecorder(object):
```
askeing: It will be helpful if we add the comments here to describe the data structure of status JSON file.

lib/common/commonUtil.py
```diff
@@ -4,27 +4,60 @@
 import platform
 import subprocess
 import numpy as np
-from datetime import datetime
+import time
```
askeing: is it possible to move up `import time` XD

askeing:
please amend the commit's message before merge it.

PR #715, opened 2017-05-23
fix browser adjustment fail when os slower

MikeLien:
1. set white color as target viewport background to compare
2. longer waiting time for checking window title exists
3. besides window title, check additional window visible attribute on windows
4. start adjustment until viewport above half setting

### DIFF
--- a/lib/baseTest.py
+++ b/lib/baseTest.py
@@ -79,9 +79,8 @@ class BaseTest(unittest.TestCase):
             viewport = dict()
             viewport_ref_fp = os.path.join(self.env.img_sample_dp, self.env.img_output_sample_1_fn)
             is_expected_viewport = False
-            time.sleep(3)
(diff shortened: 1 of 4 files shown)


MikeLien:

lib/baseTest.py
```diff
@@ -76,46 +76,46 @@ def get_browser_done(self):
         recording_enabled = CommonUtil.is_video_recording(self.firefox_config)
 
         if recording_enabled:
+            viewport = dict()
             viewport_ref_fp = os.path.join(self.env.img_sample_dp, self.env.img_output_sample_1_fn)
             is_expected_viewport = False
-            for i in range(10):
+            time.sleep(3)
```
askeing: Sleeping 3 sec without do something is wasting time.
Increasing the loop time from `20` to `25` might be the alternative solution.

PR #716, opened 2017-05-23
add README for configs

ShakoHo:
init version readme of each configs


ShakoHo:

configs/exec/README.md
```diff
@@ -0,0 +1,12 @@
+# exec configuration readme 
+  * comment : `<today>, differentiate the same build in server ` 
+  * advance : `true/false, enable will output the debug log and auto generate the mitmdump flow fliles`
```
ypwalter: * advance : `true/false, by enabling this option, it will output the debug log and generate the mitmdump flow files automatically`

configs/firefox/README.md
```diff
@@ -0,0 +1,55 @@
+# firefox configuration readme 
+  * enable_create_new_profile : `true/false, will create new firefox profile before running the whole suite`
```
ypwalter: * enable_create_new_profile : `true/false, it will create new Firefox profile before running the whole suite if enabled`

configs/chrome/README.md
```diff
@@ -0,0 +1,7 @@
+# chrome configuration readme 
+  * enable_create_new_profile : `true/false, will create new chrome profile before running the whole suite `
```
ypwalter: * enable_create_new_profile : `true/false, it will create new chrome profile before running the whole suite if enabled `

Issue #717, opened 2017-05-23
Develop the new way to inject the arguments into Sikuli from BaseTest.


askeing:
Abstract Sikuli Cases and inject the arguments.
Due to these change, we have to modify ALL existing cases.

**1st** #718
- [x] amazon IL (@askeing)
- [x] facebook IL+FT (@askeing)
- [x] gdoc IL+FT (@askeing)
- [x] gmail IL (@askeing)
- [x] gsearch IL (@askeing)
- [x] gsheet IL (@askeing)

**2nd**
- [ ] facebook PL
- [ ] gdoc PL
- [ ] gsheet PL
- [ ] gslide PL
- [ ] top sites PL

IL: Input Latency
FT: Frame Throughput
PL: Page Load

PR #737, opened 2017-05-26
Customized Region for Visual Evnets - Input Latency

askeing:
First of all, I added the region reference data into `configs/index/inputLatencyAnimationDctGenerator.json`.
This information is used for both Generator and Sikuli cases.

Second, adding the `set_override_region_settings()` methods for adding the Sikuli Override Region data if the Sikuli cases have to compare the Customized Region.
By the way, the `lib/sikuli/basecase.sikuli/basecase.py` will also reset the Sikuli Override Region data every when case init, due to the `current_status` of Stat JSON file does NOT reset automatically.

When running Sikuli case, it can find the Matched Region, which cantions the [x, y, w, h] information, and then set them into Stat JSON file by `set_override_region_settings()` method.
```python
# Customized Region
customized_region_name = 'end'
type_area = wait('type_area.png', 10)
self.set_override_region_settings(customized_region_name, type_area)
sleep(2)
```

And then I modified the `lib/converter/ftSampleConverter.py` and `lib/converter/sampleConverter.py`, because the `generate_sample_result()` method was changed from static to instance method.
Now, all methods in Generator can accese the instance variables.
Base on the above change, all methods in Generator should access `self.visual_event_points`, not `BROWSER_VISUAL_EVENT_POINTS`.
Beacuse I assign this value in `lib/generator/baseGenerator.py`.
Also, the init method will also generate the final base on the sikuli_status, which comes from Stat JSON file.
```python
# Loading Event Points settings for Customized Region
self.visual_event_points = self.BROWSER_VISUAL_EVENT_POINTS
sikuli_status = self.status_recorder.get_current_sikuli_status()
if StatusRecorder.SIKULI_KEY_REGION_OVERRIDE in sikuli_status:
    self.visual_event_points = CropRegion.generate_customized_visual_event_points(sikuli_status,
                                                                                  self.visual_event_points)
```

Finally, when we try to generate the sample result (with crop information), the `generate_sample_result_template()` method will traverse the all Sikuli Override Region data in Stat JSON file, and replace the data if needed.
The Generator will crop the following images base on the crop info of samples.


askeing:
Reference: #717

PR #738, opened 2017-05-26
clean images from outside of result run time range

MikeLien:

### DIFF
--- a/lib/generator/baseGenerator.py
+++ b/lib/generator/baseGenerator.py
@@ -167,33 +167,27 @@ class BaseGenerator(object):
             if not start_fp or not end_fp:
                 logger.warning("Cannot get file path of either start event or end event, stop clean output images.")
                 return None
+            if start_fp > end_fp:
+                logger.warning("Start point is behind End point, stop clean output images.")
+                return None
             else:
                 try:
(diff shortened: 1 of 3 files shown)


lib/generator/baseGenerator.py
```diff
@@ -167,33 +167,27 @@ def clean_output_images(self, running_time_result, img_dp):
...
-                            shutil.copyfile(imf_fp, new_img_fp)
-                    logger.debug('Remove original output image folder {}'.format(img_dp))
-                    shutil.rmtree(img_dp)
-                    logger.debug('Move temp folder {} to original output image folder {}'.format(tempdir, img_dp))
-                    shutil.move(tempdir, img_dp)
+                        start_index = max(0, files.index(start_fn) - extend_buffer)
+                        end_index = min(len(files) - 1, files.index(end_fn) + extend_buffer)
+
+                        logger.info('Cleaning {} ...'.format(root))
+                        for img_index in xrange(len(files)):
```
ypwalter: how about:

```
for img_index in range(start_index) + range(end_index+1, len(files)+1):
    img_fp = os.path.join(root, files[img_index])
    os.remove(img_fp)
```

PR #739, opened 2017-05-26
refactor gdoc and gsearch with new webapp architecture

ShakoHo:

### DIFF
--- a/lib/sikuli/common.sikuli/common.py
+++ b/lib/sikuli/common.sikuli/common.py
@@ -9,7 +9,11 @@ class General():
 
         if self.os.lower() == 'mac':
             self.control = Key.CMD
+            self.input_wheel_down_direction = WHEEL_UP
+            self.input_wheel_up_direction = WHEEL_DOWN
         else:
+            self.input_wheel_down_direction = WHEEL_DOWN
+            self.input_wheel_up_direction = WHEEL_UP
             self.control = Key.CTRL
 
     # This will take in an array of key combinations like [[Key.Enter], [Key.ENTER, Key.CTRL], ["8", Key.CTRL+Key.SHIFT]]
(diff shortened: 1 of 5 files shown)


lib/sikuli/gdoc.sikuli/gdoc.py
```diff
@@ -85,15 +159,18 @@ def page_text_generate(self, keyword, page):
...
-        click(Pattern("pics/FindReplaceInput.png").similar(0.5).targetOffset(98, 26))
+        self._click(action_name='Click replace button', similarity=0.5,
+                    component=gDoc.GDOC_CLICK_REPLACE_FINDANDPLACE_INPUT)
         type(replace_keyword)
         for i in range(replace_times):
-            wait(Pattern("pics/Replace.png").similar(0.8))
-            click(Pattern("pics/Replace.png").similar(0.8))
+            self._wait_for_loaded(component=gDoc.GDOC_REPLACE_ICON, similarity=0.8)
+            self._click(action_name='Click replace button', similarity=0.8,
+                    component=gDoc.GDOC_REPLACE_ICON)
```
ypwalter: four more spaces

lib/sikuli/common.sikuli/common.py
```diff
@@ -331,3 +339,72 @@ def il_hover(self, action_name, component, width, height, similarity=0.70, timeo
...
+        @param component: The component you want to click. ex: GMAIL_REPLY.
+        @param similarity: The similarity of image. Default: 0.70.
+        @return: true/false
+        """
+        # get the loop time base on the pattern amount of component, min loop time is 10 times
+        wait_sec = 0.5
+        loop_time = self._get_loop_times(object_amount=len(component), total_second=timeout, each_check_second=wait_sec)
+        for counter in range(loop_time):
+            for pic, offset_x, offset_y in component:
+                p = Pattern(pic).similar(similarity).targetOffset(offset_x, offset_y)
```
ypwalter: why do we need `targetOffset()` for exists?

lib/sikuli/common.sikuli/common.py
```diff
@@ -9,7 +9,11 @@ def __init__(self):
 
         if self.os.lower() == 'mac':
             self.control = Key.CMD
+            self.input_wheel_down_direction = WHEEL_UP
```
ypwalter: I don't see this get used elsewhere in General class.

lib/sikuli/common.sikuli/common.py
ShakoHo: fulfill the pattern of the image, I will leave the interface here, so if user gives the parameter 0,0, that should not affect the result?

lib/sikuli/common.sikuli/common.py
ShakoHo: for future use, in case user implements same logic again..

PR #740, opened 2017-05-29
modified bootstrap.bat and appveyor.yml for optimization

ypwalter:
1. having more cached files in appveyor
2. using most up-to-date program download link
3. make sure PATH environment variable works fine now
Verified in official windows 7 virtual machine by microsoft

It already saved 1 minutes for each appveyor run. After it started to cache files, it should save more time.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,8 +11,10 @@ skip_commits:
   message: /SkipCI.*/
 
 cache:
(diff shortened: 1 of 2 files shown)


ypwalter:
thanks.

ypwalter:
fix #633

PR #743, opened 2017-06-01
disable hint box

ypwalter:
This should be able to dismiss the most up-to-date hint box in Nightly browser.


ypwalter:
thanks

PR #745, opened 2017-06-01
bug fixing

Conjuror:


askeing:
@Conjuror well, you did not squash commits before merge O_o

Issue #748, opened 2017-06-02
Investigating the Customized Region of Input Latency & Frame Throughput cases


askeing:
List all Input Latency & Frame Throughput cases

# Amazon IL
@MikeLien #754
- [x] tests.regression.amazon.test_firefox_amazon_ail_hover_related_product_thumbnail
- [x] tests.regression.amazon.test_firefox_amazon_ail_select_search_suggestion
- [x] tests.regression.amazon.test_firefox_amazon_ail_type_in_search_field
- [x] tests.regression.amazon.test_chrome_amazon_ail_hover_related_product_thumbnail
- [x] tests.regression.amazon.test_chrome_amazon_ail_select_search_suggestion
- [x] tests.regression.amazon.test_chrome_amazon_ail_type_in_search_field

# Facebook IL
@MikeLien #766 
- [x] tests.regression.facebook.test_chrome_facebook_ail_click_close_chat_tab
- [x] tests.regression.facebook.test_chrome_facebook_ail_click_open_chat_tab
- [x] tests.regression.facebook.test_chrome_facebook_ail_click_open_chat_tab_emoji
- [x] tests.regression.facebook.test_chrome_facebook_ail_click_photo_viewer_right_arrow
- [x] tests.regression.facebook.test_firefox_facebook_ail_click_close_chat_tab
- [x] tests.regression.facebook.test_firefox_facebook_ail_click_open_chat_tab
- [x] tests.regression.facebook.test_firefox_facebook_ail_click_open_chat_tab_emoji
- [x] tests.regression.facebook.test_firefox_facebook_ail_click_photo_viewer_right_arrow

~~@ShakoHo~~ @ypwalter 
- [ ] tests.regression.facebook.test_chrome_facebook_ail_scroll_home_1_txt
- [x] tests.regression.facebook.test_chrome_facebook_ail_type_comment_1_txt
- [x] tests.regression.facebook.test_chrome_facebook_ail_type_composerbox_1_txt
- [x] tests.regression.facebook.test_chrome_facebook_ail_type_message_1_txt
- [ ] tests.regression.facebook.test_firefox_facebook_ail_scroll_home_1_txt
- [x] tests.regression.facebook.test_firefox_facebook_ail_type_comment_1_txt
- [x] tests.regression.facebook.test_firefox_facebook_ail_type_composerbox_1_txt
- [x] tests.regression.facebook.test_firefox_facebook_ail_type_message_1_txt

# Facebook FT
[...]

PR #750, opened 2017-06-02
Adding `find_match_region()` method for finding the Customized Regsion

askeing:
Sikuli cases can use this method to find the component base on multiple images (cross-platform).

Ref: #748


askeing:
@MikeLien r?

PR #755, opened 2017-06-06
use jenkins build_tag replace with build_no, then we can fix the mult…

ShakoHo:
…iple job running on the same machine problem.

### DIFF
--- a/agent/hasalTask.py
+++ b/agent/hasalTask.py
@@ -50,6 +50,9 @@ class HasalTask(object):
         self.CURRENT_WORKING_DIR = os.path.abspath(self.configurations.get('HASAL_WORKSPACE', os.getcwd()))
         self.DEFAULT_CONFIG_DP = os.path.join(self.CURRENT_WORKING_DIR, self.DEFAULT_CONFIG_DN)
 
+        # lambda function
+        self.str2bool = lambda x: x.lower() in ['true', 'yes', 'y', '1', 'ok']
+
     def update(self, **kwargs):
         if 'name' in kwargs:
             self.name = kwargs['name']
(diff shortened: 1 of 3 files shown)


agent/hasalTask.py
```diff
@@ -99,11 +98,11 @@ def create_exec_config(self):
             config_data = json.load(fh)
         config_data['max-run'] = int(self.configurations.get('MAX_RUN', 30))
         config_data['max-retry'] = int(self.configurations.get('MAX_RETRY', 15))
-        config_data['advance'] = self.configurations.get('ENABLE_ADVANCE', False).lower() == 'true'
+        config_data['advance'] = self.configurations.get('ENABLE_ADVANCE', "false").lower() == 'true'
         config_data['comment'] = self.configurations.get('EXEC_COMMENT', "<today>")
         config_data['exec-suite-fp'] = self.create_suite_file()
-        config_data['output-result-ipynb-file'] = self.configurations.get('OUTPUT_RESULT_IPYNB_FILE', False).lower() == 'true'
-        config_data['output-result-video-file'] = self.configurations.get('OUTPUT_RESULT_VIDEO_FILE', True).lower() == 'true'
+        config_data['output-result-ipynb-file'] = self.configurations.get('OUTPUT_RESULT_IPYNB_FILE', "false").lower() == 'true'
```
ypwalter: `distutils.util.strtobool()` ?

agent/hasalTask.py
```diff
@@ -79,10 +79,9 @@ def create_online_config(self):
         output_config_fp = os.path.join(self.DEFAULT_CONFIG_DP, config_dn, self.JENKINS_CONFIG_NAME)
         with open(default_config_fp) as fh:
             config_data = json.load(fh)
-        config_data['enable'] = self.configurations.get('ENABLE_ONLINE', False).lower() == 'true'
+        config_data['enable'] = self.configurations.get('ENABLE_ONLINE', "false").lower() == 'true'
```
askeing: you might want to use this function for converting string to boolean.
```python
>>> str2bool = lambda x: x.lower() in ['true', 'yes', 'y', '1', 'ok']
>>>
>>> str2bool('True')
True
>>> str2bool('false')
False
```

PR #758, opened 2017-06-06
use build flag instead of build no

ShakoHo:

### DIFF
--- a/tools/trigger_build.py
+++ b/tools/trigger_build.py
@@ -36,7 +36,6 @@ class TriggerBuild(object):
     ENV_KEY_SKIP_STATUS_CHECK = "SKIP_STATUS_CHECK"
     ENV_KEY_OUTPUT_DP = "OUTPUT_DP"
     ENV_KEY_BUILD_HASH = "BUILD_HASH"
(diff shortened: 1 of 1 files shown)


tools/trigger_build.py
```diff
@@ -90,11 +90,11 @@ def dispatch_variables(self, input_env_data):
...
 
         # assign build number to variable
-        if self.ENV_KEY_BUILD_NO in input_env_data.keys():
-            self.jenkins_build_no = input_env_data[self.ENV_KEY_BUILD_NO]
+        if self.ENV_KEY_BUILD_TAG in input_env_data.keys():
+            self.jenkins_build_tag = input_env_data[self.ENV_KEY_BUILD_TAG]
         else:
-            self.jenkins_build_no = 0
-        self.HASAL_JSON_FN = str(self.jenkins_build_no) + ".json"
+            self.jenkins_build_tag = "jenkins-unknown-0"
```
askeing: @ShakoHo how about replace `0` by timestamp?

askeing: @ShakoHo how about replace `0` by timestamp?

PR #759, opened 2017-06-06
Adding `tuning_region()` method for tuning the Region object

askeing:
It will return new Region object base on original Region object with x, y, w, h offset.


askeing:
@MikeLien r?

PR #762, opened 2017-06-07
add file exist check before rename firefox binary path

ShakoHo:
we can use this patch to avoid some machine environment not contain firefox in default path
@askeing @ypwalter @MikeLien r?

### DIFF
--- a/agent/hasalTask.py
+++ b/agent/hasalTask.py
@@ -165,7 +165,8 @@ class HasalTask(object):
                 else:
                     shutil.rmtree(firefox_fp)
         else:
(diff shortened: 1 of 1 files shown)


agent/hasalTask.py
```diff
@@ -165,7 +165,8 @@ def link_fx_pkg(self):
                 else:
                     shutil.rmtree(firefox_fp)
         else:
-            os.rename(firefox_fp, backup_path)
+            if os.path.exists(firefox_fp):
```
askeing: need some comments to describe the logic

PR #765, opened 2017-06-08
Handling the exceptions to make sure the execution of suite-teardown

askeing:


askeing:

PR #768, opened 2017-06-09
avoiding new unsent crash report notification bar

ypwalter:
There is a new unsent crash report notification bar from new firefox nightly. This should prevent us from having issues.


ypwalter:
thanks

ypwalter:
thanks

PR #769, opened 2017-06-09
Modify the `get_value_from_config` to raise warning only when value is None

askeing:
@MikeLien found there are some warning message when the value is `False`.

After tracing the code, when the value is one of `False`, `''`, `0`, and so on.
So we have to check the value, raising the warning message when it is None.


askeing:
@MikeLien r?

PR #775, opened 2017-06-12
add executor for iskakalunan

Conjuror:
this is Iskakalunan enhancement with agent.py


configs/tests/iskakalunan.json
```diff
@@ -0,0 +1,5 @@
+{
```
ShakoHo: is this the official json file for iskaklunan or just for test?

PR #776, opened 2017-06-12
fixed 3 fb cases

ypwalter:
1. modified scroll home test case so that it can create new post in homepage
2. fix pictures and code of composerbox one
3. return location for click home post area
4. comment in returned comment box

@ShakoHo @MikeLien @askeing  r?


tests/regression/facebook/test_firefox_facebook_ail_scroll_home_1_txt.py
```diff
@@ -5,6 +5,9 @@ class TestSikuli(PerfBaseTest):
 
     def setUp(self):
         super(TestSikuli, self).setUp()
+        self.set_configs(self.config_name.INDEX,
+                         self.extract_platform_dep_settings({"win32": {"7": {"compare-threshold": 0.06},
+                                                                       '10': {'compare-threshold': 0.08}}}))
```
MikeLien: threshold too large and cannot find correct result from local test on windows10 machine 6

PR #778, opened 2017-06-13
use timestamp instead of Year-Month-Day

ShakoHo:
use platform name + platform release instead of sys.platform


ShakoHo:

PR #780, opened 2017-06-13
use thirdparty tee instead of subprocess, it can help us to output lo…

ShakoHo:
…g to stdout and file at same time

need to exec python setup install to make sure this patch works


lib/thirdparty/tee.py
```diff
@@ -0,0 +1,281 @@
...
+    ... tee.system("echo test", logger=f) # output to a filehandle
+    ... tee.system("echo test", logger=print) # use the print() function for output
+    """
+    (returncode, output) = system2(cmd, cwd=cwd, logger=logger,
+                                   stdout=stdout, log_command=log_command, timing=timing, exec_env=exec_env)
+    return returncode
+
+'''
+
+class testTee(unittest.TestCase):
```
askeing: this part is test code, we might remove it.

PR #782, opened 2017-06-14
Make sure the video file is opened and ready for converting images

askeing:
After getting the `cv2.VideoCapture` object, we have to check the `isOpened()`
The `vidcap.isOpened()` will be `False` when video file is not ready.
And the `vidcap.read()` will return `(False, None)`.

Ref: #520


askeing:
@MikeLien r?

PR #786, opened 2017-06-14
update executor for real case

Conjuror:


agent/mailTask.py
```diff
@@ -197,19 +199,15 @@ def prepare_job(self, msg):
...
-               "--max-run=" + str(msg.max_run),
-               "--max-retry=" + str(msg.max_retry)]
-        if msg.calc_si != "":
-            cmd.append(msg.calc_si)
-        if msg.advance != "":
-            cmd.append(msg.advance)
-
-        logger.debug("cmd: %s" % ' '.join(cmd))
-        subprocess.call(cmd, env=os.environ.copy())
+        self.kwargs['INDEX_CONFIG_NAME'] = 'runtimeDctGenerator.json'
```
ypwalter: We do not always use 'runtimeDctGenerator.json'. Is it possible to do different json files?

Conjuror:
@ypwalter what's the expected runner for mail agent? or, is there a way to inject runner to Hasal?

ypwalter:
Recently, we use mostly `inputLatencyAnimationDctGenerator.json`
However, we do use other generators. We usually input it command line (--index-config=<str>)

Reference: `https://github.com/Mozilla-TWQA/Hasal/tree/dev/configs/index`

PR #792, opened 2017-06-15
Handling the return code of `python -m unittest`

### LINKED ISSUE #791: Handling the exceptions of `python -m unittest`
We do not handle the exception from `unittest` [1].
It seems like we have to catch the return code of `python -m unittest ...` and then raise exception when return code isn't zero for caller [2].
Then it will go to the retry part [3].

[1] https://github.com/Mozilla-TWQA/Hasal/blob/dev/runtest.py#L288
[2] https://github.com/Mozilla-TWQA/Hasal/blob/dev/runtest.py#L266
[3] https://github.com/Mozilla-TWQA/Hasal/blob/dev/runtest.py#L271-L276

askeing:
fixed #791


askeing:

PR #793, opened 2017-06-15
Modify the GMail type-for-reply-btn images, find type area base on that

askeing:
After getting the type-for-reply-btn, we shift x-axis right 12~13 px, and add the height 50 px to get the type area.


askeing:
@MikeLien @ShakoHo @ypwalter r?

PR #799, opened 2017-06-16
modify for get revert timestamp for video usage

MikeLien:
**WIP**

We record timestamps for video initial and actions from t1 and t2 for following analysis and calculation like below:
```
-------o-----------o--------o------
    initial       t1       t2        (timestamps)
```

But since we don't know the exactly start point for video start (currently we wait video file exist for [ffmpeg recording](https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/profiler/avconvProfiler.py#L29-L34) and [obs recording](https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/profiler/obsProfiler.py#L163-L174)), we may record incorrect initial timestamp while computer running slower.

A new idea is that we assume video recording termination will faster then starting video recording, thus we use another time stamp as below:
```
initial
|------------------o--------o------o--
                  t1       t2     end  (timestamps)
```

We can use total frames of video and recorded timestamps to calculate the duration from actual `initial time` to `t1` and `t2` and it might be much precise.

### DIFF
--- a/configs/index/runtimeDctGenerator.json
+++ b/configs/index/runtimeDctGenerator.json
@@ -16,17 +16,20 @@
   "platform-dep-settings"   : {
     "darwin":{
       "default": {
-        "video-recording-fps": 60
+        "video-recording-fps": 60,
+        "search-margin": 10
       }
     },
     "win32":{
       "default": {
-        "video-recording-fps": 60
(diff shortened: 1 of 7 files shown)


MikeLien:
r?

askeing:
lgtm

PR #813, opened 2017-06-27
flush pending log before sending target message

MikeLien:


lib/sikuli/common.sikuli/common.py
```diff
@@ -100,6 +109,7 @@ def set_mouse_delay(self, sec=0.5):
     def set_type_delay(self, sec=0):
         Settings.TypeDelay = sec
 
+    @FlushBeforePrint
```
ShakoHo: you should set this decorator before _screenshot_and_time instead of system_print...then you don't need to add the system_print in front of _screenshot_and_time everytime...

PR #815, opened 2017-07-10
RC 1.2.0

ShakoHo:
@MarkYan Please help to verify this PR in Win7 and Win10 (each platform should at least run IL case once)


MarkYan:
Test succeed in both platforms.

MarkYan:
@ShakoHo Test succeed in both platforms.

PR #819, opened 2017-07-12
add cmd path in config

ShakoHo:

### DIFF
--- a/configs/chrome/default.json
+++ b/configs/chrome/default.json
@@ -5,5 +5,22 @@
     "Cookies"             : "Cookies_gsuite_fb_linda",
     "Secure Preferences"  : "Secure Preferences",
     "Preferences"         : "Preferences"
+  },
+  "platform-dep-settings"   : {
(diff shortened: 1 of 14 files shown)


askeing:
it seems like we also have to change the `firefoxProfileCreator.py` and `chromeProfileCreator.py`

- https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/helper/firefoxProfileCreator.py#L16-L19
- https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/helper/chromeProfileCreator.py#L19-L27

askeing:
it seems like we also have to change the `firefoxProfileCreator.py` and `chromeProfileCreator.py`

- https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/helper/firefoxProfileCreator.py#L16-L19
- https://github.com/Mozilla-TWQA/Hasal/blob/dev/lib/helper/chromeProfileCreator.py#L19-L27

PR #820, opened 2017-07-13
fix incorrect usage of decoration

MikeLien:


lib/sikuli/common.sikuli/common.py
```diff
@@ -385,7 +385,7 @@ def _screenshot_and_time(self, width, height,
         current_time = time.time()
 
         # Print log message for recording
-        self.common.system_print(action_name)
+        General().system_print(action_name)
```
askeing: due to we did not use `self`, it seems like we can change `_screenshot_and_time()` from instance method to static method.

PR #821, opened 2017-07-13
add facebook type 100 txt in composerbox cases

MikeLien:


lib/sikuli/facebook.sikuli/facebook.py
```diff
@@ -427,17 +427,17 @@ def click_post_area_home(self, type='center'):
         @param type:
         @return:
         """
-        location = None
+        location = pattern = obj = None
```
ypwalter: I think this will work fine.

However, it is somehow dangerous when it is not assigning `None`.
See `https://stackoverflow.com/questions/16348815/python-assigning-multiple-variables-to-same-value-list-behavior/16349356#16349356`
Maybe try ... `location, pattern, obj = [None, None, None]`?

@askeing thoughts?

ypwalter:
Maybe we should merge all N ms test cases into 1 test cases with multiple entry points or config files?

tests/regression/facebook/test_chrome_facebook_ft_type_composerbox_100_txt_delay_30ms.sikuli/test_chrome_facebook_ft_type_composerbox_100_txt_delay_30ms.py
```diff
@@ -0,0 +1,69 @@
...
+        sleep(2)
+        _, _, obj = fb.click_post_area_home()
+        sample1_fp = os.path.join(self.INPUT_IMG_SAMPLE_DIR_PATH, self.INPUT_IMG_OUTPUT_SAMPLE_1_NAME)
+        sample2_fp = os.path.join(self.INPUT_IMG_SAMPLE_DIR_PATH, self.INPUT_IMG_OUTPUT_SAMPLE_1_NAME.replace('sample_1', 'sample_2'))
+
+        sleep(2)
+        capture_width = int(self.INPUT_RECORD_WIDTH)
+        capture_height = int(self.INPUT_RECORD_HEIGHT)
+
+        com.set_type_delay(0.01)
```
ypwalter: a comment here would be even better

PR #824, opened 2017-07-17
Disable onboarding notification. 

ypwalter:
It will crash Nightly if in classic theme. This will diable onboarding notificaitons.


ypwalter:
thanks

ypwalter:
Note: even though this patch fixed most related issues, but there might be some computer with weird behavior. In that case, please do:

A. Purge Firefox
1. about:support
2. Refresh Firefox

B. Clean Up Default Firefox Profile
1. `Firefox -P` in command line
2. remove old profile and create a new one

PR #836, opened 2017-07-25
Fix the runtest config validation

askeing:


askeing:
@ShakoHo @MikeLien r?

PR #837, opened 2017-07-25
fix write message encode with utf-8

ShakoHo:
@askeing r?

### DIFF
--- a/lib/thirdparty/tee.py
+++ b/lib/thirdparty/tee.py
@@ -84,8 +84,16 @@ def system2(cmd, cwd=None, logger=_sentinel, stdout=_sentinel, log_command=_sent
             msg += '\n'
             try:
                 f.write(msg)
-            except TypeError:
-                f.write(msg.encode("utf-8"))
+            except Exception:
+                try:
+                    write_ctnt = msg.encode("utf-8", "ignore")
+                except Exception:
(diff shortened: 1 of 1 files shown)


lib/thirdparty/tee.py
```diff
@@ -84,8 +84,16 @@ def filelogger(msg):
             msg += '\n'
             try:
                 f.write(msg)
-            except TypeError:
-                f.write(msg.encode("utf-8"))
+            except Exception:
+                try:
+                    write_ctnt = msg.encode("utf-8", "ignore")
+                except Exception:
+                    print "WARNING: encode utf-8 error, message will be skipped!!!"
```
askeing: using `print('foo')` to replace `print 'foo'`

lib/thirdparty/tee.py
```diff
@@ -84,8 +84,16 @@ def filelogger(msg):
...
+                    write_ctnt = ""
+                try:
+                    f.write(write_ctnt)
+                except Exception:
+                    print "write to stdout with utf-8 error"
```
askeing: as above

lib/thirdparty/tee.py
```diff
@@ -95,6 +103,8 @@ def filelogger(msg):
             sys.exit(-1)
         pass
 
+
+
```
askeing: I will prefer only keep two empty line between two methods.

PR #842, opened 2017-07-27
Fix Firefox Profiler Creation

ypwalter:
make Firefox command quoted and it would raise exception when there is something wrong.

### DIFF
--- a/lib/helper/firefoxProfileCreator.py
+++ b/lib/helper/firefoxProfileCreator.py
@@ -4,6 +4,7 @@ import time
 import json
 import shutil
 import tempfile
+import subprocess
 from lib.common.logConfig import get_logger
 from lib.common.pyDriveUtil import PyDriveUtil
 from ..common.environment import Environment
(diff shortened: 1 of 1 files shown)


lib/helper/firefoxProfileCreator.py
```diff
@@ -36,10 +36,18 @@ def get_firefox_profile(self, prefs={}, cookies_settings={}, profile_files_setti
 
     def _create_firefox_profile(self):
         tmp_profile_dir = tempfile.mkdtemp(prefix='firefoxprofile_')
+        quoted_firefox_cmd = '"%s"' % self.firefox_cmd.strip()
+
         logger.info('Creating Profile: {}'.format(tmp_profile_dir))
-        os.system('{} --profile {} -silent'.format(self.firefox_cmd, tmp_profile_dir))
-        self._firefox_profile_path = tmp_profile_dir
-        logger.info('Creating Profile success: {}'.format(self._firefox_profile_path))
+        status = os.system('{} --profile {} -silent'.format(quoted_firefox_cmd, tmp_profile_dir))
```
askeing: how about put Line.39 and Line.42 into one line
```python
status = os.system('"{}" --profile {} -silent'.format(self.firefox_cmd.strip(), tmp_profile_dir))
```

askeing: if we replace `os.system` by `subprocess.check_call`, then it will run command and wait for command to complete. If the return code was not zero then raise `CalledProcessError`.

ypwalter: OK, let me try it out first.

ypwalter:
tested in local machine and it's working.
@askeing r? again

ypwalter:
Thanks.

PR #844, opened 2017-07-27
changed the cmd so that it works in mac and linux

ypwalter:
This is to make sure cmd works in mac and linux, too.
@ShakoHo, please help verify if this works in your mac.

r?


ypwalter:
thanks.

PR #853, opened 2017-08-03
add amazon and facebook type 100 character case with 100ms interval

MikeLien:
1. add amazon type 100 character case
2. add facebook type 100 character case
3. fix incorrect amazon type case with 50ms interval setting
4. enlarge facebook comparison region


MikeLien:

PR #865, opened 2017-08-07
Adding Pulse "mgt" topic for sending Agent response to Trigger

askeing:
- modify the time interval of checking in `trigger_config.json` from 1 min to 10 min.
- modify `pulse.py` to check username/pwd before listen Pulse.
- add `response_agent_info()` in `pulse.py` to response Agent info to Pulse `mgt` topic, so Trigger can get the Agent info.
- modify `tasksTrigger.py` to check username/pwd, raise Exception if there is no account info.
- add `job_listen_response_from_agent()` in `tasksTrigger.py` to listen the `mgt` topic for getting Agent info from Agents and save them into Rotating Logger.
  - `TimedRotatingFileHandler`: interval 1 day, 30 backup files will be kept.

### DIFF
--- a/configs/ejenti/trigger_config.json
+++ b/configs/ejenti/trigger_config.json
@@ -6,7 +6,7 @@
       "enable": false,
       "topic": "win7",
       "platform_build": "win64",
-      "interval_minutes": 1,
+      "interval_minutes": 10,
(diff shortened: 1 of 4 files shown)


askeing:

PR #867, opened 2017-08-08
fix fb find contact error

MikeLien:


MarkYan:
With this patch, I successfully run these cases on my local computers. @MikeLien 
Thanks for fixing.

PR #874, opened 2017-08-10
change the order of upload data registration

ypwalter:


ypwalter:
#876 fixed the same issue. closed this pr as it is.

PR #876, opened 2017-08-10
remove hasal server dependency

ShakoHo:


configs/upload/default.json
```diff
@@ -0,0 +1,11 @@
+{
+  "enable"                      : false,
+  "perfherder_protocol": "http",
```
MarkYan: perfherder-protocol

lib/generator/baseGenerator.py
```diff
@@ -291,13 +292,13 @@ def output_suite_result(global_config, index_config, exec_config, output_result_
         except Exception as e:
             logger.error('Cannot move statistics file to result folder. Error: {exp}'.format(exp=e))
 
-        # copy current result file to result folder
+        # move current result file to result folder
```
ypwalter: why do you want to use "move" instead?

lib/helper/uploadResultHelper.py
```diff
@@ -0,0 +1,343 @@
...
+                        }
+                    ],
+                    # List of job guids that were coalesced to this job
+                    'coalesced': []
+                }
+            }
+        ]
+
+    def create_job_collection(self, dataset):
+        tjc = TreeherderJobCollection()
```
ypwalter: maybe adding a comment to usage page `https://github.com/mozilla/treeherder/blob/master/docs/submitting_data.rst`

runtest.py
```diff
@@ -232,24 +235,94 @@ def get_test_env(self, **kwargs):
...
+        if self.upload_config['perfherder-pkg-platform']:
+            result['PERFHERDER_PKG_PLATFORM'] = str(self.upload_config['perfherder-pkg-platform'])
         else:
             result['PERFHERDER_PKG_PLATFORM'] = ""
         for variable_name in kwargs.keys():
             result[variable_name] = str(kwargs[variable_name])
         return result
 
-    def run_test_result_analyzer(self, test_case_module_name, test_name, current_run, current_retry, video_result):
+    def upload_test_result_handler(self):
```
ypwalter: This is a comment for me to come back and look it from this point. XD

lib/generator/baseGenerator.py
ShakoHo: running the same test in next round will affect the test result if we don't move it to the result folder

runtest.py
```diff
@@ -232,24 +235,96 @@ def get_test_env(self, **kwargs):
...
-        result['ONLINE_CONFIG_FP'] = self.online_config_fp
+        result['UPLOAD_CONFIG_FP'] = self.upload_config_fp
         result['SUITE_RESULT_DP'] = str(self.suite_result_dp)
         result['FIREFOX_PROFILE_PATH'] = self._firefox_profile_path
         result['CHROME_PROFILE_PATH'] = self._chrome_profile_path
 
-        if self.online_config['perfherder-revision']:
-            result['PERFHERDER_REVISION'] = str(self.online_config['perfherder-revision'])
+        if self.upload_config['perfherder-revision']:
+            result['PERFHERDER_REVISION'] = str(self.upload_config['perfherder-revision'])
```
askeing: in `get_test_env()` method, we should raise Exception if there is no `perfherder-revision`, `perfherder-pkg-platform`.

MarkYan:
You might also change Hasal/tools/generate_data_template.json,
whose "perfherder_suite_name" should be "perfherder-suitename".
@ShakoHo

PR #879, opened 2017-08-14
Refactoring `check_pulse_queue_exists()` and add job exception listener

askeing:
Also modify the rotating interval to `midnight`, due to the rollover time of TimedRotatingFileHandler is computed from log file's `mtime`.


askeing:

PR #881, opened 2017-08-14
Fix the `hasalTask` configs parameters missing issue

askeing:
Patch comes from @ShakoHo 
Ref: https://gist.github.com/ShakoHo/04049978371fa8ae0955287490f6c99a


askeing:
@ShakoHo r?

PR #936, opened 2017-08-18
change typing 100 chars case to type lorem ipsum txt

MikeLien:

### DIFF
--- a/tests/regression/amazon/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.sikuli/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.py
+++ b/tests/regression/amazon/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.sikuli/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.py
@@ -58,9 +58,9 @@ class Case(basecase.SikuliInputLatencyCase):
         t1 = time.time()
         capimg1 = capture(0, 0, capture_width, capture_height)
 
-        char_len = 100
-        sample_str = string.letters + string.letters[::-1]
-        char_str = (sample_str * (char_len / len(sample_str) + 1))[:char_len]
+        # Reference from https://en.wikipedia.org/wiki/Lorem_ipsum
+        # extract 100 chars from Lorem ipsum text and filter characters which will affect result, e.g., i, j, and l
+        char_str = "orempsumdorstametconsecteturadpscngetsedoeusmodtemporncduntutaboretdoremagnaquatenmadmnvenamqusnoexe"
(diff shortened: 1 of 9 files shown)


tests/regression/amazon/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.sikuli/test_chrome_amazon_ft_type_100_txt_in_search_field_delay_100ms.py
```diff
@@ -58,9 +57,9 @@ def run(self):
         t1 = time.time()
         capimg1 = capture(0, 0, capture_width, capture_height)
 
-        char_len = 100
-        sample_str = string.letters + string.letters[::-1]
-        char_str = (sample_str * (char_len / len(sample_str) + 1))[:char_len]
+        # Reference from https://en.wikipedia.org/wiki/Lorem_ipsum
+        # extract 100 chars from Lorem ipsum text and filter characters which will affect result, e.g., i, j, and l
+        char_str = "orempsumdorstametconsecteturadpscngetsedoeusmodtemporncduntutaboretdoremagnaquatenmadmnvenamqusnoexe"
```
ShakoHo: IIRC, Kanru suggests we should use meaningful word and sentence instead of random non-meaningful strings.

MikeLien: **A common form of lorem ipsum reads:**
`Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.`

As discussed before, we should avoid `i`, `j`, `l`, upper case characters, etc. to minimize noise to impact comparison result. The string which I use comes from its original Lorem ipsum form but remove characters we don't want. If we need choose meaningful word, we can try to find `some words` from Lorem ipsum text in English version and without noise characters. @ShakoHo Thought?

ShakoHo: can we find a meaningful sentence without i,j,l from Lorem ipsum?

PR #939, opened 2017-08-18
Refactoring the cases of amazon IL hover related product thumbnail

askeing:
Refactoring `new Region` by `tuning_region()` method


askeing:
@ShakoHo @MikeLien r?

PR #942, opened 2017-08-21
fix bug: time transfer to local time

MarkYan:
Date and time returned by querying Perfherder are in UTC. However, our reference data and time are in UTC+8, which is local time of Taiwan. Therefore, this patch modified the input time to UTC+8.


MarkYan:
@ypwalter r?

tools/backfill_agent.py
```diff
@@ -167,23 +173,27 @@ def create_json(self, refill_date='latest'):
                         fout.write('\"{}\"'.format(refill_date))
 
                 elif '\"TEMPLATE\": \"ADD_SUITE_HERE\"' in line:
-                    # always random select 5 cases to run
-                    case_count = 5
-                    if case_count < len(self.backfill_list):
-                        ran_select = random.sample(self.backfill_list, case_count)
+                    if refill_date == 'latest':
```
ypwalter: can be simplified to 1 if-else

```
                    case_count = 5
                    # In latest mode, random select 5 cases to run.
                    # In history mode, run all suite
                    if refill_date == 'latest' and case_count < len(self.backfill_list):
                        selected_suites = random.sample(self.backfill_list, case_count)
                    else:
                        selected_suites = self.backfill_list
```

PR #947, opened 2017-08-24
modified path accordingly so that it could work again

ypwalter:
In this commit, I made some changes so that more windows 10 machines might be able to use this bootstrap script.

```
1. detect folder of Conda and clean up related code
2. detect jre8 path
```

@ShakoHo @askeing @MikeLien r?

It was tested in hp windows10 machines and worked well.


ypwalter:
thanks

PR #948, opened 2017-08-24
	Reset `current_status` before each case run

askeing:
The `loop_suite()` will call `loop_test()` for each cases, and then `loop_test()` will set the case information into `running_statistics` file.

We should reset `current_status` of each case run, so we add `clean_legacy_status()` and call it before `run_test()`.
Or previous cases' customized region will affect next case.


askeing:
@ypwalter r?

PR #949, opened 2017-08-24
modified google search response image

ypwalter:
@askeing r?


askeing:
Thanks @MarkYan 's verification, it works fine on ESR45.

askeing:
Thx :)

Send from my iPhone

PR #953, opened 2017-08-29
Handle Slack RTM message for leader and agent

askeing:
Adding Slack feature, leader commands and sync/async commands


askeing:

PR #958, opened 2017-08-31
Facebook scrolling down case

ypwalter:
@ShakoHo @MikeLien @askeing 
Please review the case for me.

I don't think this will be merged into our branches. However, for better diff and comment, I still do a pr here. Please help me with reviewing if this is  a valid test. It already finished its first run in Windows 7 (30 times). If you would like to see the customized region and result for this, please come to my desk for it.

Thanks.


ypwalter:
Per discussion last week, I think I will still need to check this code in. Please review it for me.
1. A new test case for Firefox only
2. A new ipynb
3. The image will not be in the library since this is for SC's whitehat account only.

*Test results are on the azure server
https://notebooks.azure.com/ypwalter/libraries/ipynb
Result of off-main-thread disabled
Result of off-main-thread enabled

*Running images and videos uploading is still in progress, I expected it to be finished by end of this weekend
https://drive.google.com/open?id=0BwDTW9d_sBJMSWxqUzBmTDNsejQ

MikeLien:
Based on better management, I'll suggest to make another folder for requested cases.

ypwalter:
The scripts was moved to a newly created tests group "request"

PR #959, opened 2017-08-31
Adding the ipynb template for creating multiple data source report

askeing:
Since there are more requests for generating the merged reports, let's writing a template for it.

User can modify the value of `HIDE_MISSING_CASE` and `INPUT_SOURCE` to generate their report.


askeing:

PR #972, opened 2017-09-08
change folder uri for account switch

ShakoHo:
the new client cert files are already uploaded to google drive, after this PR is merged, all agent machines need to download the new certs files
@askeing , @MikeLien @ypwalter r?


askeing:
## **Important**

"after this PR is merged, all agent machines need to download the new certs files"

PR #979, opened 2017-09-13
Skip reporting error if there are some None stat in running_statistics file

askeing:
The stat will be `None` when Hasal is still running, so we should skip it when we get `None` stat.

error example, we should skip it:
**[Agent]** DESKTOP-????????/10.*.*.*
**[ERROR]** We found some error occurred from running_statistics file.
**[Case Name]** test_firefox_gslide_ail_type_0
**[Timestamp]** 20170913161521
**[Sikuli Running Stat]** None
**[Image Comapre Stat]** None
**[FPS Stat]** None
**[Time]** 2017-09-13 16:16:01


askeing:

PR #986, opened 2017-09-14
Modify the default logging level to Info

askeing:


askeing:
After having the discussion with team members, we should change the default logging level to INFO

PR #989, opened 2017-09-14
Modify logging level of running_statistics_monitor

askeing:
Base on @ShakoHo suggestion, sending WARN when file cannot open.


askeing:
@ShakoHo @MikeLien @ypwalter

PR #990, opened 2017-09-15
fix compare threshold and region for gsuite

ShakoHo:


ShakoHo:

PR #994, opened 2017-09-18
fix access denied when rename downloaded fx package

ShakoHo:
modify the default latest backfill to 14 days


tools/backfill_agent.py
```diff
@@ -224,7 +224,7 @@ def run_history(self, backfill_config):
         # TODO: add file checker
 
         # Assume back fill dates are in 3 months
```
ypwalter: comment should:
1. 3 months -> 14 days
2. Explain that 14 days is for dashboard

PR #1001, opened 2017-09-20
Modify the timezone issue of `ref_date` for backfill_agent.py

askeing:
We should also modify Hasal Dashboard again.
The reason is we have to find correct ref_date first, then calcuate the `value_ds` and `count_ds`.
Ref: https://github.com/Mozilla-TWQA/Hasal_dashboard/pull/1


ycyen:
r+, thank you for the fix!

PR #1002, opened 2017-09-20
Adding clean_pulse_queues feature for pulse_trigger with `--clean`

askeing:
Re-create the Pulse queue for cleaning Dead Consumer Client on Pulse.
Dead Consumer Client will get messages without ack(), so messages will always stay on Pulse, and no one can handle it.

### DIFF
--- a/ejenti/pulse_modules/hasalPulsePublisher.py
+++ b/ejenti/pulse_modules/hasalPulsePublisher.py
@@ -1,3 +1,4 @@
+import time
 import pickle
 import logging
 from mozillapulse.messages.base import GenericMessage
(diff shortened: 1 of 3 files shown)


askeing:

ejenti/pulse_modules/hasalPulsePublisher.py
```diff
@@ -83,6 +84,37 @@ def got_msg(body, message):
...
+            # wait for deleting ...
+            for _ in range(10):
+                queue_exist = c.queue_exists()
+                if not queue_exist:
+                    break
+                time.sleep(1)
+            logging.info('Queue Topic [{}] delete ... {}'.format(topic, 'OK' if not queue_exist else 'Failed'))
+
+        # re-create Queue
+        return HasalPulsePublisher.check_pulse_queue_exists(username=username, password=password, topic=topic)
```
MikeLien: Here will always get `True` except encountering `Exception`
method `check_pulse_queue_exists` doesn't return exactly queue exists or not.

askeing: Modify the return value of `check_pulse_queue_exists()` method.
https://github.com/Mozilla-TWQA/Hasal/pull/1002/files#diff-e4dad583837c5cb9acd5c085df9a94daR85

askeing:
@MikeLien r? again

PR #1003, opened 2017-09-21
Do NOT remove md5 folder if it exists and is folder, only remove it if it is file

askeing:


askeing:
@MikeLien r?

ejenti/pulse_modules/tasksTrigger.py
```diff
@@ -252,10 +248,7 @@ def clean_md5_by_job_name(job_name):
...
         # prepare MD5 folder
         if os.path.exists(md5_folder):
-            if os.path.isdir(md5_folder):
-                shutil.rmtree(md5_folder)
-                os.makedirs(md5_folder)
-            elif os.path.isfile(md5_folder):
+            if os.path.isfile(md5_folder):
                 os.remove(md5_folder)
         else:
             os.makedirs(md5_folder)
```
MikeLien: Depends on method functionality, it could `return` after `makedirs` because it's an empty folder and doesn't need to clean anything.

ejenti/pulse_modules/tasksTrigger.py
```diff
@@ -202,10 +201,7 @@ def check_latest_info_json_md5_changed(job_name, platform):
...
```
MikeLien: Depends on method functionality, it could `return` after `makedirs` because it's an empty folder and doesn't need to check any change.

askeing:
@MikeLien @ypwalter @ShakoHo r?

PR #1007, opened 2017-09-22
The `input_cmd_str` is not only reference field, it will be used for determining the interactive mode or not

askeing:


askeing:

PR #1008, opened 2017-09-22
Modify pulse_trigger configs template file

askeing:


askeing:
it will fail if running `run-hasal-on-latest-nightly` command without following fields...
- `CHECKOUT_LATEST_CODE_REMOTE_URL`
- `CHECKOUT_LATEST_CODE_BRANCH_NAME`

PR #1011, opened 2017-09-22
Adding tasks for querying and removing Hasal Output folder

askeing:


askeing:

PR #1018, opened 2017-10-09
disable DRM notification

ypwalter:
We encountered an issue that a notification UI on top of Firefox browser is showing and blocking us from detecting the targeted waiting element. Even if I disable DRM using `media.eme.enabled=false`, it would still show me an message for asking me to reactivate DRM.

In that case, I made `browser.eme.ui.enabled=false`, and this should disable any further message from Firefox about DRM media related information.

@ShakoHo @askeing r?

### DIFF
--- a/configs/firefox/default.json
+++ b/configs/firefox/default.json
@@ -35,7 +35,8 @@
     "browser.onboarding.enabled": false,
     "browser.onboarding.hidden": true,
     "browser.onboarding.notification.finished": true,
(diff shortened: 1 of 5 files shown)


ypwalter:
Tried in win7 asus machines. It works

ypwalter:
Also, firefox pagedown pattern was changed. There is a 2nd commit about one `Key.UP` added for it.

PR #1019, opened 2017-10-11
Adding machine name, which contains hostname and IP address

askeing:
We upload performance data to Perfherder with machine name `local-machine` currently.
We should replace magic string by hostname and IP address, so we can find which data came from which machine.


askeing:
@ShakoHo @ypwalter @MikeLien r?

askeing:
It works fine on Stage, https://treeherder.allizom.org/#/jobs?repo=mozilla-central&selectedJob=129551830

PR #1020, opened 2017-10-12
modified amazon hover area for new ff pagedn pattern

ypwalter:
1. Firefox text style changes from time to time
2. There are 2 "customer" text between the desired screenshot area. We changed to locate "Product details" and move the (x,y) back to original designed position.

@ShakoHo @askeing r?
@ycyen, can you verify this for me?


ycyen:
Verified on Firefox and Chrome on Win7 and Win10!

ypwalter:
Thanks!

PR #1021, opened 2017-10-13
Adding wait 1~3 sec for load-balance

askeing:
Base on @ypwalter 's observation, some pulse clients will take more tasks than others.
Try to add the wait time for load-balance.


askeing:
@ShakoHo @ypwalter r?

PR #1022, opened 2017-10-16
modified fb home post area for the new UI (20171016)

ypwalter:
The new Facebook UI changed its wording and action order. This would adding two new pictures for the UI to be recognized.

@ShakoHo @askeing r?


ypwalter:
thanks

PR #1028, opened 2017-10-18
Adding ejenti and tools into epydoc.cfg

askeing:
will add `ejenti` and `tools` into http://mozilla-twqa.github.io/Hasal/api/


askeing:
@ShakoHo @ypwalter r?

ShakoHo (approved):

PR #1041, opened 2017-11-02
Using common library for mask the confidential values, and write unitest for it

askeing:


askeing:
@ShakoHo @ypwalter r?

ShakoHo (approved):

PR #1047, opened 2017-11-14
modify perfherder with multi signatures

ShakoHo:
@askeing @ypwalter r?


ejenti/pulse_modules/tasksTrigger.py
```diff
@@ -31,6 +31,8 @@ class TasksTrigger(object):
     KEY_CONFIG_PULSE_USER = 'pulse_username'
     KEY_CONFIG_PULSE_PWD = 'pulse_password'
 
+    KEY_CONFIG_SIGNATURE_WHITE_LIST = 'signature_white_list'
```
askeing: should be `suite_white_list_for_signature`

ejenti/pulse_modules/tasksTrigger.py
```diff
@@ -80,6 +82,7 @@ def __init__(self, config, cmd_config_obj, clean_at_begin=False):
         self.jobs_config = config.get(TasksTrigger.KEY_CONFIG_JOBS, {})
         self.pulse_username = config.get(TasksTrigger.KEY_CONFIG_PULSE_USER)
         self.pulse_password = config.get(TasksTrigger.KEY_CONFIG_PULSE_PWD)
+        self.signature_white_list = config.get(TasksTrigger.KEY_CONFIG_SIGNATURE_WHITE_LIST, [])
```
askeing: modify template config for reference should be better

configs/ejenti/trigger_config.schema
```diff
@@ -74,7 +78,8 @@
     "pulse_username",
     "pulse_password",
     "jobs",
-    "log_filter"
+    "log_filter",
+    "suite_white_list_for_signature"
```
askeing: Due to we assign the default value by following code
```python
self.suite_white_list_for_signature = config.get(TasksTrigger.KEY_CONFIG_SUITE_WHITE_LIST_FOR_SIGNATURE, [])
```
It seems like we don't have to set `suite_white_list_for_signature` as `required` field.

askeing (approved):
only one comment for `required`.

lgtm

PR #1050, opened 2017-11-15
update new icon for gslide

ycyen:
@ypwalter r?


ypwalter:
The order can be changed as we discussed offline. The upper the more recent image should be in the list. Thanks and approved.

ypwalter (approved):
r+

PR #1051, opened 2017-11-15
Modify `setup.py` for clean up some files and extend ejenti requirements

askeing:


askeing:
@ShakoHo @ypwalter r?

setup.py
```diff
@@ -12,11 +15,34 @@ def validate_pywin32():
...
+# Clean up some critical files which will crash the program
+#
+clean_file_list = [
+    os.path.join('ejenti', 'ejenti.pyc'),
+]
+
+for clean_file in clean_file_list:
+    print('Checking {} ... '.format(clean_file), end='')
+    if os.path.exists(clean_file):
+        print('found')
```
ypwalter, line 31: I think `print("Found" + clean_file)` or `print('Found {} ... '.format(clean_file), end='')` could be better, and it should be the same for L39.

askeing: `print("Found" + clean_file)` will become
```bash
Checking FOO ... Found FOO
```

`print('Found {} ... '.format(clean_file), end='')` will become
```bash
Checking FOO ... Found FOO ... 
```

ypwalter: Sorry, I didn't read it carefully. :P

ypwalter (approved):
r+

PR #1052, opened 2017-11-15
Fill the value list when there is no case result when generating the dashboard data

askeing:
When there is no result on Perfherder, the `backfill_table` will have no record in `perfherder_data`.

For exampe, when we have three cases **foo_case**, **bar_case**, and **latest_case**.
And the **bar_case** has no result:
```json
"perfherder_data": {
      "foo_case:firefox:windows8-64": {
        "value": [
          33.33,
          22.22
        ],
        "signature": "foo_sig"
      },
      "latest_case:firefox:windows8-64": {
        "value": [
          44.44
        ],
        "signature": "latest_sig"
      }
}
```

That's the reason why we have to check all keys, and then fill no result keys into `self.source_data`.


askeing:
@ShakoHo @ypwalter r?

ShakoHo (approved):
lgtm r+

PR #1054, opened 2017-11-17
Adding Back Fill Hasal Result feature into Trigger Server base on Builds

askeing:


askeing:
@ShakoHo @ypwalter r?

ejenti/server_jobs/backfill_trigger.py
```diff
@@ -0,0 +1,234 @@
...
+        StatusFileCreator.create_status_file(job_id_fp, StatusFileCreator.STATUS_TAG_PULSE_TRIGGER_BACKFILL, 100)
+
+        # check queue
+        queue_exists = HasalPulsePublisher.check_pulse_queue_exists(username=self.pulse_username,
+                                                                    password=self.pulse_password,
+                                                                    topic=topic)
+
+        if not queue_exists:
+            logging.error('There is not Queue for Topic [{topic}]. Message might be ignored.'.format(topic=topic))
+
```
ShakoHo, line 60: if the queue does not exist, do we need to continue the following steps?

ShakoHo (approved):

ejenti/server_jobs/backfill_trigger.py
askeing: It's a trade-off between cpu resource and case running.
Not only publisher can create queue, any one of consumers (agents) also can create queue.
It still has chance to submit tasks after queue be created by consumer.

ejenti/server_jobs/backfill_trigger.py
```diff
@@ -0,0 +1,234 @@
...
+from lib.modules.build_information import BuildInformation
+from lib.common.commonUtil import CommonUtil
+from ejenti.pulse_modules.hasalPulsePublisher import HasalPulsePublisher
+
+
+class BackFillTrigger(object):
+
+    PLATFORM = 'win64'
+    RESULT_AMOUNT = 6
+    WAIT_TIMEOUT_DELTA_HOUR = 8
```
askeing: it seems like I should move this parameter out to config file

PR #1060, opened 2017-11-23
clone status content before manipulate it

ShakoHo:
@ypwalter r?
- add more comments
- clone status content before manipulate it


ypwalter (approved):
r+

PR #1062, opened 2017-11-27
fix ref config affection

ShakoHo:
@ypwalter r?


ypwalter (approved):
r+
This pr is a trial and might not be able to fix the problem. However, this will not affect current logic.

PR #1064, opened 2017-11-28
Sometimes the overwrite_cmd_configs will be overwrite by other methods, deepcopy it to avoid it

askeing:
ref: #1062


askeing:
@ShakoHo @ypwalter r?

ShakoHo (approved):

